PR #12, opened 2018-02-05
add RESOURCES.md

zackschuster:
only a few things added, mainly to kick things off. i'm sure there's a bajillion resources that could/should be added.

no real overriding principle here besides "this is related to nodejs esm work" & "i think making it more accessible would help people"

### DIFF
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -1,3 +1,10 @@
+# Resources
+
+## A collection of items related to Node.js module work
+
+### Reading
 * [The Current State of Implementation and Planning for ESModules](https://medium.com/the-node-js-collection/the-current-state-of-implementation-and-planning-for-esmodules-a4ecb2aac07a)
 * [ESM modules in node: npm edition](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054)
+
+### Meetings
 * [Meeting availability](https://doodle.com/poll/cqavkzwxtxzccs4z)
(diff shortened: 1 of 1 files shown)


RESOURCES.md
```diff
@@ -0,0 +1,3 @@
+* [The Current State of Implementation and Planning for ESModules](https://medium.com/the-node-js-collection/the-current-state-of-implementation-and-planning-for-esmodules-a4ecb2aac07a)
```
MylesBorins: can you add a title and maybe a bit more context?

WebReflection:
Maybe worth considering these links too:

  * [cjs4esm](https://github.com/WebReflection/cjs4esm#cjs4esm--) brings CommonJS modules and `require` to ESM compatible environments such [SpiderMonkey](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey), [JSC](https://trac.webkit.org/wiki/JavaScriptCore), or even browsers
  * [CGJS](https://github.com/cgjs/cgjs#cgjs-) is a CommonJS runtime for [GJS](https://wiki.gnome.org/Projects/Gjs)
  * [import.js](https://github.com/WebReflection/import.js#importjs) is a dynamic `import(...)` polyfill
  * SpiderMonkey [ESM Loader](https://searchfox.org/mozilla-central/source/js/src/shell/ModuleLoader.js)

There's quite possibly tons of code to share for modules there, I hope some resource here makes sense.

RESOURCES.md
zackschuster: done :)

zackschuster:
would that be under a "prior art"-type section? sounds like it could be extremely helpful :smile: 

i'd like to just get the base doc merged for now tho if that's ok? i feel like it would make everything go smoother overall :)

MylesBorins:
I'm going to land this as is. @WebReflection do you want to submit a follow u PR?

PR #14, opened 2018-02-06
Add Prior Art Section

WebReflection:
As suggested in PR #12 it could be interesting
to investigate or share code and knowledge
about CommonJS in ESM and/or vice-versa

### DIFF
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -8,3 +8,31 @@
 
 ### Meetings
 * [Meeting availability](https://doodle.com/poll/cqavkzwxtxzccs4z)
+
+### Prior Art
+
+Following a list of ESM and CommonJS relevant projects.
+
+#### Transpilers
+
+  * [Babel](http://babeljs.io) transpiled ES6+ including modules downgraded to CommonJS with an `exports.default` strategy
+  * [TypeScript](https://www.typescriptlang.org) transpiled classes including ES6 like/inspired modules
+
+#### Bundlers
+
+  * [Webpack](https://webpack.js.org) multi purpose bundler compatible with CommonJS `exports.default` strategy. Actually diverging from npm resolution preferring a `browser` field in the `package.json` instead of actual CommonJS.
+  * [Rollup](https://rollupjs.org) next-generation ES6 module bundler with tree shaking capability
+  * [Parceljs](https://parceljs.org/) multi purpose bundler focused on zero config and performance
+  * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
+
+#### Compilers
+
+  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
+  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
+
+#### Runtime Utilities
+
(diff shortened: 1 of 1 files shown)


benjamn:
Would https://github.com/benjamn/reify and https://github.com/standard-things/esm fall into this category?

WebReflection:
mind providing same format w/ quick description per project?

benjamn:
Added some more details, hopefully not too much!

WebReflection:
you wrote `@std/esm` as link instead of @std/esm (backtics)

WebReflection:
thanks @benjamn, we overlapped a little bit so I've cleaned up all changes into a single squashed commit

jdalton:
Hi! I just `watch`ed this repo so I should get notifications on activity. Thanks for the ping in @benjamn via https://github.com/nodejs/modules/commit/4d0385c4b6c8a36e6674f03a93f1d3ffd8e29f0d. If we're to help differentiate between `@std/esm` and reify it would be:
> `@std/esm` is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.

WebReflection:
@jdalton done

WebReflection:
maybe I miss something but what has Babel to do with modules?

bmeck:
@WebReflection I'm not sure we should exclude things just because they solve multiple use cases.

MylesBorins:
@WebReflection I think that babel is useful to include as many modules were authored based on the way the module loader worked (shim of require)

bmeck:
We should probably also consider Webpack since it also implements a module loader entirely separate from Babel if we include babel. Traceur might also be related since it has a way of compiling down. Are there any AMD based solutions that I'm missing (I'm not as well versed in AMD solutions)?

DanielRosenwasser:
TypeScript also provides interop strategy that basically treats a `module.exports` and namespace object as 1:1. It also more or less implements the Babel/Webpack strategy as of 2.7.

WebReflection:
> We should probably also consider Webpack ...

that's what I was talking about ... then rollup, all my offline and runtime utilities, then parcel ... then ...

bmeck:
@WebReflection yup, all are prior art. I have plenty of tools as well.

benjamn:
Babel and TypeScript are both used by server-side applications in Node, whereas client-side bundlers like Webpack are less relevant for this working group, since it's not common to run a Webpack bundle in Node, so there's not much Node-specific insight to take from Webpack.

weswigham:
@WebReflection In many ways, es modules as presented by [babel](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/), [webpack](https://webpack.js.org/concepts/modules/), [rollup](https://github.com/rollup/rollup) and [systemjs](https://github.com/systemjs/systemjs), or [typescript](https://www.typescriptlang.org/docs/handbook/modules.html), are some of the most widely used forms of them in use today. It'd be a big omission to elide them. Plus, it's a list of links, what can it hurt.

WebReflection:
I actually agree with @benjamn and I have no budget (in terms of time) to change this PR per each person that wrote a browser side module solver ... can we stick with just Babel and TypeScript as more Node relevant?

DanielRosenwasser:
> so there's not much Node-specific insight to take from Webpack.

Webpack provides an interoperability strategy to consume CommonJS modules with ES module syntax, and specially operates on ES-module-authored CommonJS modules, much like the other tools here. so I'm not sure why you wouldn't want to account for it.

bmeck:
@WebReflection I request Webpack not be avoided, it is one of the largest tools used by the node ecosystem and also is able to [target Node](https://webpack.js.org/concepts/targets/#usage).

WebReflection:
Split in sections with all the famous bundlers included.

WebReflection:
I think Myles is busy now, I'll do it since there are 2 approvals.

PR #26, opened 2018-02-14
Document quorum required in order to merge PRs

bmeck:
Per: https://docs.google.com/document/d/17T8ZeF2mUPffiDvulqd_HqD_JrTNaf8ch2dFT1hMfD0/edit?disco=AAAABp_P_B0

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -62,7 +62,11 @@ Note that some changes to Team membership require unanimous consensus.  See
 
 ### Merging PRs into this Repository
 
-Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository. Pull requests that fail to be merged will be addressed on a case by case basis.
+
+#### Errata and Editorial Changes
+
+A special exception is made for pull requests seeking to fix errata or perform editorial changes to this repository. Errata or editorial pull requests may be merged without being presented at a meeting if a reasonable time is given for review and there no dissent. The time period seen as reasonable for review varies on a case by case basis as determined by the author. A member may request a specific time period for review of a errata or editorial pull request not to exceed the next meeting date. If a time for review is requested, members must wait for that time period to pass or review be completed prior to that time.
 
 <a id="developers-certificate-of-origin"></a>
 ## Developer's Certificate of Origin 1.1
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -62,7 +62,7 @@ Note that some changes to Team membership require unanimous consensus.  See
 
 ### Merging PRs into this Repository
 
-Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
 
 <a id="developers-certificate-of-origin"></a>
 ## Developer's Certificate of Origin 1.1
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -60,6 +60,10 @@ or else the discussion will continue. Simple majority wins.
 Note that some changes to Team membership require unanimous consensus.  See
 "Team Membership" above.
 
+### Merging PRs into this Repository
+
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
+
 <a id="developers-certificate-of-origin"></a>
 ## Developer's Certificate of Origin 1.1
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -62,6 +62,8 @@ Note that some changes to Team membership require unanimous consensus.  See
 
 ### Merging PRs into this Repository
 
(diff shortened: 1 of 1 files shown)


WebReflection:
My only concern is that working on something and then not reaching the quorum after working, maybe a lot, might not be good to incentive contributions.

Could the possibility to reach the quorum upfront be also considered? I mean as a ticket with an issue/improvement instead then once the PR landed. Does this make any sense?

bmeck:
@WebReflection I think we could make an exception for bug fixes or editorial tweaks, but probably not a general case of approving a PR before the PR is finished.

WebReflection:
fair enough

GOVERNANCE.md
```diff
@@ -60,6 +60,10 @@ or else the discussion will continue. Simple majority wins.
 Note that some changes to Team membership require unanimous consensus.  See
 "Team Membership" above.
 
+### Merging PRs into this Repository
+
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
```
MylesBorins: We should likely document that there needs to be approval in the PR and that dissenting in the PR is also sufficient to block something landing

---

Also

s/quorum must not be merged into this repository./quorum may not be merged into this repository.

bmeck: @MylesBorins "may" generally implies that it is optional in [IETF terms](https://www.ietf.org/rfc/rfc2119.txt). I think "must" is the right word here.

Do we need pre-approval in the PR before bringing it to a committee? Also we didn't specify what kind of dissent is blocking and if the committee cannot override that dissent. Maybe we could leave that for later / as needed.

bmeck:
@ljharb since we talked about observers in the meeting and were noting that observers were not affecting quorum I assume they would not be counted.

MylesBorins:
@ljharb would you like to open a follow up PR defining `Observer`? If not I'm more than happy to take that on.

MylesBorins:
Should we include a note about fast tracking doc fixes or errata?

benjamingr:
I'm +1 on including said note - I don't think we should wait 2 weeks for things that don't constitute of actual design decisions.

giltayar:
LGTM

manekinekko:
LGTM

GOVERNANCE.md
```diff
@@ -60,6 +60,10 @@ or else the discussion will continue. Simple majority wins.
 Note that some changes to Team membership require unanimous consensus.  See
 "Team Membership" above.
 
+### Merging PRs into this Repository
+
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
```
manekinekko: LGTM.

What happens to the PRs that are not merged? Do they got simply closed?

benjamingr: @manekinekko I think this can be decided on a case-by-case basis

bmeck: Called out as being addressed on case by case basis.

bmeck:
@MylesBorins @benjamingr I have added the errata and editorial exception.

GOVERNANCE.md
```diff
@@ -62,7 +62,11 @@ Note that some changes to Team membership require unanimous consensus.  See
 
 ### Merging PRs into this Repository
 
-Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository.
+Pull requests must reach consensus in a meeting in order to be merged into this repository. Merging of pull requests must be decided during a meeting that has a minimal quorum of 50% of all active members. Members with observer status are not considered active members for the purposes of reaching quorum. A pull request that is is unable to reach quorum due to lack of consensus or quorum must not be merged into this repository. Pull requests that fail to be merged will be addressed on a case by case basis.
```
jasnell: I want this to be clear: this requirement applies only to PRs affecting *this repository* correct? This requirement should not apply to PRs impacting ESM code in the main @nodejs/node repository.

Even with this repository, I'm fairly -1 on the language here. Merging of a pull request should require only that there are sign-offs and no objections within a reasonable period of time. Requiring that we must have a meeting with a minimal majority to make decisions is going to force everything into a vote every time, which is simply not necessary.

bmeck: @jasnell 

> Pull requests must reach consensus in a meeting in order to be merged into this repository.

I think I was explicit that this applies only to this repository. If there is a better wording we can change the wording.

The meeting requirements was decided during the previous meeting. We can discuss at the next one and wait for that before merging this PR but this was the decision during a meeting.

jasnell: > I think I was explicit that this applies only to this repository

A statement specifically indicating that the rule does not apply to ESM module related code changes in the main repo would be sufficient, I think. It's strictly about setting expectations, really. I don't want a situation where some bit of code gets merged per the nodejs/node rules and have folks argue that some process defined here wasn't followed... if that makes sense.

jasnell: > The meeting requirements was decided during the previous meetings...

Yes, I was there. I believe this part needs more consideration.

bmeck: @jasnell added note about scope of this section.

> Yes, I was there. I believe this part needs more consideration.

We can hold of for now pending the next meeting. However, I would not be comfortable merging this with a different decision prior to then.

giltayar: I must admit that I am uncomfortable with this decision too. I deferred to the group's decision in the group, as I am new to all this, but I believe that deferring most PRs to the meeting is a bit too much process for me. It will mean lots of PRs waiting in the queue.

Having "exceptions" to this process also sounds like a problem, as we now need to define what constitutes an exception to this rule, both in a formal document, and probably as part of some of the PRs that want a "fast track".

I say let's give 72 hours to object to a PR. It's a good compromise between waiting two weeks for a meeting, and the regular 48 hours. Maybe even 96 hours.

But, given that I'm new to this, if @jasnell removes his objection, and everyone's good with it, then so am I.

MylesBorins:
@bmeck do you think we should move this into a `CONTRIBUTING.md`

GOVERNANCE.md
```diff
@@ -60,6 +60,16 @@ or else the discussion will continue. Simple majority wins.
 Note that some changes to Team membership require unanimous consensus.  See
 "Team Membership" above.
 
+### Merging PRs into this Repository
+
+This section does not apply to [the Node.js core repository](https://github.com/nodejs/node). It only applied to [the Modules repository](https://github.com/nodejs/modules).
```
MylesBorins: s/applied/applies

bmeck: fixed

bmeck:
@MylesBorins we can put it wherever, but it only is telling how members should act not how contributions should be made. I think a link might be better than moving this talk about how members act into a CONTRIBUTING document that focuses on how to makea PR. -0 to moving it, but +0 on linking from contributing guidelines.

GOVERNANCE.md
targos: forgot to push?

PR #30, opened 2018-02-16
Corrected categorization and language for RESOURCES.md

DanielRosenwasser:
Mostly

* ~transpiler -> compiler~
* aligned the phrasing between Babel & TypeScript (they both support the same interop mode)
* I generally consider Reify and `@std/esm` to be closer to runtime tools than compilers.

Let me know if there are any objections here.

### DIFF
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -20,9 +20,9 @@ Following a list of ESM and CommonJS relevant projects.
   * [Parceljs](https://parceljs.org/) multi purpose bundler focused on zero config and performance
   * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
 
-#### Compilers
-  * [Babel](http://babeljs.io) compiled ES2015+ including modules downgraded to CommonJS
-  * [TypeScript](https://www.typescriptlang.org) compiled ES2015+ including modules downgraded to CommonJS
+#### Transpilers
+  * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
+  * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
   
 #### Runtime Utilities
   * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -24,9 +24,11 @@ Following a list of ESM and CommonJS relevant projects.
   * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
   * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
   
-#### Runtime Utilities
+#### Loaders
   * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
   * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
+
+#### Runtime Utilities
   * [cjs4esm](https://github.com/WebReflection/cjs4esm#cjs4esm--) brings CommonJS modules and `require` to ESM compatible environments such [SpiderMonkey](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey), [JSC](https://trac.webkit.org/wiki/JavaScriptCore), or ESM compatible browsers
   * [import.js](https://github.com/WebReflection/import.js#importjs) is a dynamic `import(...)` polyfill
   * [SpiderMonkey](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey) [ESM Loader](https://searchfox.org/mozilla-central/source/js/src/shell/ModuleLoader.js)
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -14,10 +14,6 @@
 
 Following a list of ESM and CommonJS relevant projects.
 
-#### Transpilers
-  * [Babel](http://babeljs.io) transpiled ES6+ including modules downgraded to CommonJS with an `exports.default` strategy
-  * [TypeScript](https://www.typescriptlang.org) transpiled classes including ES6 like/inspired modules
-
 #### Bundlers
   * [Webpack](https://webpack.js.org) multi purpose bundler compatible with CommonJS `exports.default` strategy. Actually diverging from npm resolution preferring a `browser` field in the `package.json` instead of actual CommonJS.
   * [Rollup](https://rollupjs.org) next-generation ES6 module bundler with tree shaking capability
@@ -25,10 +21,12 @@ Following a list of ESM and CommonJS relevant projects.
   * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
 
 #### Compilers
-  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
-  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
-
+  * [Babel](http://babeljs.io) compiled ES2015+ including modules downgraded to CommonJS
+  * [TypeScript](https://www.typescriptlang.org) compiled ES2015+ including modules downgraded to CommonJS
+  
 #### Runtime Utilities
+  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
+  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
   * [cjs4esm](https://github.com/WebReflection/cjs4esm#cjs4esm--) brings CommonJS modules and `require` to ESM compatible environments such [SpiderMonkey](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey), [JSC](https://trac.webkit.org/wiki/JavaScriptCore), or ESM compatible browsers
   * [import.js](https://github.com/WebReflection/import.js#importjs) is a dynamic `import(...)` polyfill
   * [SpiderMonkey](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey) [ESM Loader](https://searchfox.org/mozilla-central/source/js/src/shell/ModuleLoader.js)
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -22,8 +22,8 @@ Following a list of ESM and CommonJS relevant projects.
 
 #### Transpilers
   * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
-  * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
-  
+  * [TypeScript](https://www.typescriptlang.org) TypeScript is a typed superset of JavaScript that compiles to plain JavaScript, including a CommonJS interop layer with ESM
+
 #### Loaders
   * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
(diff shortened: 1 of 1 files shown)


RESOURCES.md
```diff
@@ -14,21 +14,19 @@
...
   * [Webpack](https://webpack.js.org) multi purpose bundler compatible with CommonJS `exports.default` strategy. Actually diverging from npm resolution preferring a `browser` field in the `package.json` instead of actual CommonJS.
   * [Rollup](https://rollupjs.org) next-generation ES6 module bundler with tree shaking capability
   * [Parceljs](https://parceljs.org/) multi purpose bundler focused on zero config and performance
   * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
 
 #### Compilers
-  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
-  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
-
+  * [Babel](http://babeljs.io) compiled ES2015+ including modules downgraded to CommonJS
```
weswigham: > downgraded 
Perhaps "converted"? I'm pretty sure there's no strict ranking relationship between the two formats, other than maybe a temporal one.

streamich:
Why `transpiler -> compiler`?

weswigham:
> compiler - a program that converts instructions into a machine-code or lower-level form so that they can be read and executed by a computer.

Both tools lower ES6+ syntax to older versions of the spec, making them a compiler. "transpiler" is JS community jargon ;)

streamich:
So, you are saying that transpiling one version of a language to another version of the same language is compiling?

RESOURCES.md
```diff
@@ -14,21 +14,19 @@
...
-  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
-  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
-
+#### Transpilers
+  * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
+  * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
+  
 #### Runtime Utilities
+  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
+  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
```
ljharb: Both of these are more “loaders”, perhaps?

DanielRosenwasser:
Done.

RESOURCES.md
```diff
@@ -14,21 +14,19 @@
...
   * [Parceljs](https://parceljs.org/) multi purpose bundler focused on zero config and performance
   * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
 
-#### Compilers
-  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
-  * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
-
+#### Transpilers
+  * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
+  * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
```
ljharb: TypeScript is a different language; it doesn’t convert primarily/just ES2015+ as i understand it. (eg, TS supports a number of keywords that at this juncture are never likely to land in the language)

MylesBorins:
I've pinged @jdalton (std/esm) and @benjamn (reify) for review

RESOURCES.md
weswigham: It also processes plain ol' es6+ and downlevels it, just like Babel.

ljharb: Fair. But i think it’s important to call out that that’s not it’s primary purpose.

DanielRosenwasser:
Switched to loaders.

RESOURCES.md
ljharb: Considering it broke the ecosystem module interop between versions 1.8 and 2.7 (iirc, the synthetic imports thing), I’d say it’s pretty relevant to this repo that TypeScript is a different language.

weswigham: > Considering it broke the ecosystem module interop between versions 1.8 and 2.7 (iirc, the synthetic imports thing)

? What do you mean?

ljharb: I believe (i don’t use TS myself, but this is what TS users have led me to believe was the case) that by default, during these versions, you had to import CJS modules using `import * as`, which would then yield a spec-violating value instead of anything close to a proper ModuleRecord.

weswigham: Here's a bit of a story: When es modules were first implemented in Babel and TS, neither project had ground to stand on with regards to how interop with cjs was going to work. TS initially decided that, absent any direction, it'd take what it was told by a definition file at face value (note how almost every TS declaration file is written using module syntax) and just translate an import statement into a require statement (since, ofc, the syntax prevents any shenanigans). Now, that alone still didn't allow non-conforming imports (at least in the type system); but some clever users realized they could make a non-conforming import typecheck if they merged a namespace and class or function and assigned that to the module's export object (which is a collection of individual abilities added to make typing existing cjs libraries possible), since TS was using namespaces to check for module-ness. By the time TS started noticing this (and went to close the hole in the typesystem), the community had started to take a dependency on it, and TS didn't have a better answer for how cjs aught to be importable in ESM (and note, it still doesn't). When we see people abusing it, we're always quick to say that it's not real es6 and that if you need that kind of interop you really should use the nonstandard `require`-style import to flag the usage as interop import. Users then typically complain that other tools can't handle those import forms for bundling/whathaveyou (man, I love when everything is all in on a someplaces underspecified someplaces overspecified standard) and continue using breaking behaviors. Some time passes, and Babel's style of interop becomes popular. TS doesn't really like it, since it's also not specified, standardized, guaranteed to be forward compatible, or stable, but to bow to the greater popularity of Babel, it starts emitting the esModule marker on it's emit to be compatable with Babel's import helper, and adds the allowSyntheticDefaultImports flag to let you assume that cjs imports will be available on the default during typechecking (still not changing the emit to interop in any way, believing that what's happened so far is untenable in the future, but allowing you to correctly typecheck a babel-transpiled codebase). Fast forward another _two years_, and there's still no finalized plan for cjs-esm interop in node, but Babel project interop is honestly more important at the current point in time (esm users: near zero. Babel users: many). To allow users to use TS and Babel more interchangeably, TS adds a flag that causes us to emit similar interop helpers to Babel. (Hoping beyond hope that since the community's rallied around this interop scheme that it essentially becomes the standard and we can unflag it.) And that's where we are today. TS's original stance was and has always been that there's no defined way for the two systems to interop at present, but it has given concessions to community desires over time to interop with other transpilers/ecosystems, at least. You can call it conditionally non-conforming, sure, but when there's no official consensus on interop to begin with there's no ground to stand on. And by your own admission it does have an optionally "conforming" implementation today, so I'm not even seeing what your objection actually is. And even then, a non-conforming implementation would be just a educational as a confirming one, since hopefully you'd be able to examine the design tradeoffs and acquire more data on the relative value of them.

This is a bit of a tangent, but tbqh I'm tired of the nitpicking of all the details instead of just getting an admission that "hey, that project also has an esm implementation that made slightly different design decisions than some other projects over time and maybe the history of their implementation can be a useful learning resource for us". There was never a "let's use ESM syntax but not" decision in the TS team - consistently TS has tried to hew as close to the standards as the community and ecosystem lets it (and usually enforcing or at least recommending even more). And that _should_ be a lesson to everyone here; _don't mess this up_, because ultimately the seven people currently implementing this and the 40 people in this working group and their opinions are _dwarfed_ by the silent majority in the ecosystem who will really decide how this feature gets used and its relative success. Dictating from on high how things should be has seemed terribly ineffective in the JS community, since that seems to just cause proliferation of more tools to make things work how community members think it should.

ljharb: Thanks for the history! I'm not trying to malign anybody here, and it's fine to think of TS's choices as simply a different design decision.

That said, I continue to believe that it's necessary to call out that TS is not JavaScript, it's a different language that transpiles to JavaScript - the current wording implies that it's solely/primarily a babel alternative, which I hope you'll agree is not its primary use case.

weswigham: > That said, I continue to believe that it's necessary to call out that TS is not JavaScript, it's a different language that transpiles to JavaScript - the current wording implies that it's solely/primarily a babel alternative, which I hope you'll agree is not its primary use case.

It might not be primary, but it's an important one. The TS team probably invests close to as much time in ecosystem compat and transpiler development as type system development, and it's not like it doesn't support parts of JS - all JS is valid TS, since TS is supposed to be a superset of JS that layers on types (the non-type additions are akin to babel stage 0 stuff that never advanced that TS has to keeps for language stability reasons). And it's a bit disingenuous to claim babel's any different, considering it does very little on it's own (it coordinates plugins - right on it's page about plugins it says "Now, out of the box Babel doesn’t do anything"), and you can get plugins to make it many different hundreds of flavors of not-JS-that-transpile-to-JS. Unless ofc you'd like to limit the scope of "primary use" to a specific plugin, at which point I'd decry that instead of external plugins TS just has all the plugins builtin and compiler options to switch them on and off  - and it's definitely the primary use case of the "module" flag 😉

ljharb: Sure, it can continue to say what it currently does; I just want it to accurately describe TypeScript, which I don't think the current text does.

weswigham: None of these descriptions really describe what the thing is beyond the broad category, just how they use/relate to modules.

RESOURCES.md
```diff
@@ -14,18 +14,18 @@
...
   * [Webpack](https://webpack.js.org) multi purpose bundler compatible with CommonJS `exports.default` strategy. Actually diverging from npm resolution preferring a `browser` field in the `package.json` instead of actual CommonJS.
   * [Rollup](https://rollupjs.org) next-generation ES6 module bundler with tree shaking capability
   * [Parceljs](https://parceljs.org/) multi purpose bundler focused on zero config and performance
   * [Browserify](http://browserify.org) the CommonJS module bundler that brought NodeJS to the Web
 
-#### Compilers
-  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS, supporting live bindings without rewriting imported variable names
+#### Transpilers
+  * [Babel](http://babeljs.io) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
+  * [TypeScript](https://www.typescriptlang.org) converts ES2015+ to older versions, including a CommonJS interop layer with ESM
```
ljharb: my earlier request about terminology is resolved, but i still think that implying that TypeScript only/primarily takes JS as an input is deceptive, and i'd like to see some tweak to indicate this.

jdalton: I rather leave the TS description to the TS maintainers in the WG. Sorry you don't agree with how the project maintainers describe their project.

ljharb: The link describes their project as "TypeScript is a typed superset of JavaScript that compiles to plain JavaScript". I'm fine with that description.

RESOURCES.md
```diff
@@ -14,18 +14,18 @@
...
+  
+#### Loaders
+  * [Reify](https://www.npmjs.com/package/reify) compiles ESM syntax to CommonJS on the fly, supporting live bindings without rewriting imported variable names
   * [@std/esm](https://github.com/standard-things/esm) is a fork of Reify focused on following Node's ESM support roadmap while providing a bridge from the CJS/ESM usage of today to the ESM of tomorrow.
```
jdalton: `@std/esm` is now `esm` 🎉.

MylesBorins:
Removing from the agenda as this does not yet have consensus in the PR

MylesBorins:
Is this something we want to continue to try and land or should we close it?

MylesBorins:
ping one more time on if this should remain open

GeoffreyBooth:
I say land it.

GeoffreyBooth:
@ljharb Can I ask you to either push an update to change it or to just let it go? This has been open all year. It’s time to move on.

ljharb:
@GeoffreyBooth generally those interested in seeing something land are the ones that have the obligation to push a change. "Time to move on" would mean closing this and not landing it.

That said, sure, I can update the description.

weswigham:
The copy is verbatim the tagline from our website now, so I can't say it's wrong - though I'd disagree with the opinion that the old copy was wrong, too. 🤷‍♀️ 

It's fine - I doubt anyone will have trouble finding us or actually knowing what we are. Tbh, our biggest usage is actually as the static analysis engine driving the most popular IDEs for both JS and TS, but I guess that's not something we really advertise~

MylesBorins:
This has been opened forever and has no objections I can see... I'm landing

Issue #8, opened 2018-02-05
Governance and Membership Requirements


MylesBorins:
If this team plans to become a chartered working group we will need to determine a governance model as well as membership requirements. I'll use this original post to track the discussion. Currently listing the high level things that need to be determined

## Governance Model

* Decision Making Model

## Membership Requirements

* Participation Rules

benjamingr:
I vote for consensus based decision making? Seems to be working well elsewhere so far.

targos:
What should happen if consensus cannot be reached?

giltayar:
And, pardon my ignorance, how do we know consensus is reached?

targos:
@giltayar As long as there are no objections (within a reasonable amount of time), we usually consider there is consensus.

benjamingr:
@giltayar sorry, I forgot some of us aren't collaborators (yet!) - generally the same way Node does decision making in core. 

 1. Someone raises a question or issue or makes a PR.
 2. Other members debate it for a reasonable amount of time (typically 48h or 72h on weekends)
 3. If everyone is in agreement and no one raises an objection - it gets merged.
 4. Objections are debated until everyone is in agreement (or in agreement to not block the change).
 5. If consensus cannot be reached (often it can with more or more concise discussion) , voting happens. In this case this would either be the team members or the TSC.

In practice, _most_ people agree on _most_ issues consensus seeking has worked in the past. Everyone can voice their concerns and are working towards the same goal. I know modules has been painful for some participants in the past - but I believe the consensus seeking model can work here. 

See also https://github.com/nodejs/TSC/blob/master/TSC-Charter.md#section-8-voting for what node does.

WebReflection:
> As long as there are no objections
> If everyone is in agreement

AFAIK consensus is different from unanimity, this should be clear to every contributor (or we gonna have hard time doing any change at all).

> Objections are debated until everyone is in agreement (or in agreement to not block the change)

I'd explicitly write it in the second way:

> Objections are debated until everyone is in agreement to not block the change

If explicit, we can point at it. If it's a sub-clause, it's harder to reason about it.

Again, in favour of consensus but with those points clarified and easy to point at in every discussion whenever if needed. I hope this is reasonable.

giltayar:
[...]

Issue #3, opened 2018-02-01
Convert to Working Group through Chartering


MylesBorins:
Currently this group is set up as a "team", it would be way better if we were to be chartered. We can use this issue to track process. I will periodically update this original comment to have the most up to date summary of the current state of the process.

mcollina:
I’m +1 to chartering this and giving full autonomy.

mhdawson:
No objection.  Past approach has usually been for the new group to meet for a while and then request to be chartered.

mcollina:
I think removing people after 1-2 weeks of inactivity is against the Node.js values.

weswigham:
@mcollina we don't even have a documented governance process yet; I don't think we've decided that that's going to be a thing yet, and it'd be very appropriate to raise such a concern both in the governance issue (as is now raised), and at the next meeting during the discussion about that issue.

mcollina:
@weswigham from https://github.com/nodejs/modules/pull/31: 

> As was discussed in the last meeting all members that did not fill out
the availability doodle have been moved to observer status.

That was decided in the previous meeting, and this is what I am referring about.
This _has_ been decided. I'm reacting to that decision.

MylesBorins:
I'm going to close this for now. We can reopen the issue when we are thinking about moving forward with this.

PR #31, opened 2018-02-18
doc: use ncu-team sync to sync member list

MylesBorins:
As was discussed in the last meeting all members that did not fill out
the availability doodle have been moved to observer status.

Observers are not counted toward quorum and do not get to vote

Refs: https://github.com/nodejs/modules/issues/8#issuecomment-364261695
Refs: https://github.com/nodejs/modules/pull/26/

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,41 +17,50 @@ Work includes:
 
 <!-- ncu-team-sync.team(nodejs/modules) -->
 
-- [@abouthiroppy](https://github.com/abouthiroppy) - Yuta Hiroto
 - [@benjamingr](https://github.com/benjamingr) - Benjamin Gruenbaum
 - [@benjamn](https://github.com/benjamn) - Ben Newman
 - [@bmeck](https://github.com/bmeck) - Bradley Meck
-- [@bmeurer](https://github.com/bmeurer) - Benedikt Meurer
 - [@ceejbot](https://github.com/ceejbot) - C J Silverio
 - [@chrisdickinson](https://github.com/chrisdickinson) - Chris Dickinson
 - [@DanielRosenwasser](https://github.com/DanielRosenwasser) - Daniel Rosenwasser
-- [@devamaz](https://github.com/devamaz) - Ahmad Abdul-Aziz 
+- [@dduleone](https://github.com/dduleone) - Dan DuLeone
+- [@devamaz](https://github.com/devamaz) - Ahmad Abdul-Aziz
 - [@devsnek](https://github.com/devsnek) - Gus Caplan
-- [@eugeneo](https://github.com/eugeneo) - Eugene Ostroukhov
-- [@evanplaice](https://github.com/evanplaice) - Evan Plaice
 - [@Fishrock123](https://github.com/Fishrock123) - Jeremiah Senkpiel
 - [@giltayar](https://github.com/giltayar) - Gil Tayar
 - [@guybedford](https://github.com/guybedford) - Guy Bedford
+- [@iarna](https://github.com/iarna) - Rebecca Turner
 - [@inidaname](https://github.com/inidaname) - Hassan Sani
 - [@jasnell](https://github.com/jasnell) - James M Snell
 - [@jdalton](https://github.com/jdalton) - John-David Dalton
 - [@jkrems](https://github.com/jkrems) - Jan Olaf Krems
+- [@justinfagnani](https://github.com/justinfagnani) - Justin Fagnani
 - [@linclark](https://github.com/linclark) - Lin Clark
 - [@ljharb](https://github.com/ljharb) - Jordan Harband
 - [@manekinekko](https://github.com/manekinekko) - Wassim Chegham
-- [@mcollina](https://github.com/mcollina) - Matteo Collina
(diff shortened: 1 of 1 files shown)


mcollina:
I’m -1 for obvious reasons. I was traveling most of this week, and I _thought_ I filled in that doodle. The fact that people would have been kicked out was not written in the text of an issue, just in a single comment and maybe in meeting notes. Considering the amount of messages this group generates in a week, it was very easy to miss.

We are typically very sensitive about these things in this organization, so I encourage everyone to reconsider this approach.

joyeecheung:
By the way: you can create child teams under @nodejs/modules , and put observers in some where like @nodejs/modules-observers, non-observers in @nodejs/modules-members (with write access to this repo), so you can still ping everyone with the parent team, and sync the lists using ncu-team, and when there is a PR you can tell the difference by the greeness of their sign-offs.

joyeecheung:
Also...@nodejs/automation may help developing a tool that count votes by a specific team, now come to think of it.

benjamingr:
> I’m -1 for obvious reasons. I was traveling most of this week, and I thought I filled in that doodle. The fact that people would have been kicked out was not written in the text of an issue, just in a single comment and maybe in meeting notes. Considering the amount of messages this group generates in a week, it was very easy to miss.

There were at least 5 pings about this - and Myles reached out to everyone who hasn't filled out the doodle by email (I think). That said, I think the idea was mainly to not have people who didn't have time to participate in the team given the voting process and consensus model.

I'm personally fine with a rule like "observers who participate in two straight meetings and in issues are moved back to member status" and "members who don't participate in two meetings or the issue tracker are moved to observer status".

weswigham:
@benjamingr Maybe add your proposal to #8 so we remember to consider it in the next meeting?

mcollina:
Let me be very straightforward: Myles ping me only right before this PR was filed via twitter DM. I received no other direct ping.

And I just thought the doodle in question was https://t.co/TWYtjwpE5y. My mistake.

jdalton:
@mcollina The PR isn't merged (and even if so) if there was a misunderstanding I'm sure things could get ironed out. The "take my toys and go home" tone isn't really needed here. No one is against you. No one is out to get you. The doodle bit was confusing. It almost tripped up Sean too. No worries.

mcollina:
> The PR isn't merged (and even if so) if there was a misunderstanding I'm sure things could get ironed out. 

This PR has already been acted upon, as I'm no longer part of the group. It was not a question, but rather a representation of reality. Making it a question would have been enough.

> The "take my toys and go home" tone isn't really needed though. No one is against you. No one is out to get you.

Absolutely! I never thought that was the case. I was upset, and my tone was a bit over where it should have been.

> The doodle bit was confusing (it almost tripped up Sean too). 

When membership is considered, there should be nothing to "trip". It should be crystal clear if someone is in or out. People take time off, go on vacation, and might need to take some "away" time because of burnout. __We should not decide on a Wednesday that someone has untill Sarturday to fill something or they are out.__ This sets a bad precedent for this group, and the whole Node.js project. We can do better.
This is what I'm referring about in https://github.com/nodejs/modules/issues/3.

> I say no worries if you want on the team members list.

The group can decide that no, a one week period is not enough to remove a member. Someone might be off for a month _at least_. Then, ask them if they are still committed and would like to step down voluntarily, and after some time remove them if there is no answer. Everyone should have the chance to make that decision voluntarily first.
This is my proposal.

benjamingr:
@mcollina would you perhaps be interested in being a member of this team again and the decision on the removal discussed in the next meeting?

mcollina:
@benjamingr I care that people are allowed enough personal time away before being removed, and that should be asked directly several times. I would probably join back in some time if this passes. I still think this group work is critical for our community.

MylesBorins:
In retrospect, and based on comments in this thread and https://github.com/nodejs/modules/issues/8 I would like to suggest that we revisit the decision to move the eight members removed by this PR. This is already tagged for discussion in the upcoming meeting, but perhaps we can work out some of the ideas in #8 and put the findings in this PR before we discuss on the 28th.

@nodejs/modules thoughts?

@joyeecheung I've gone ahead and refactored the groups so that we have a modules team, with two subteams. Everyone in this PR who is currently moved to observers are in the observers subteam. With the permission of @mcollina I've added him to that team and back out the earlier commit from  @benjamingr 

Apologies again for how this turned out, especially on a long weekend.

abouthiroppy:
@MylesBorins 

I apologize for the late reply.
Should I join Node.js Foundation Modules Team Meeting? I can't join because it will be held at 4 AM at Tokyo.

README.md
```diff
@@ -17,41 +17,50 @@ Work includes:
 
 <!-- ncu-team-sync.team(nodejs/modules) -->
```
joyeecheung: You can change this to `<!-- ncu-team-sync.team(nodejs/modules-collaborators) -->`, and use another `<!-- ncu-team-sync.team(nodejs/modules-observers) -->` block in the next section. `ncu-team` would sync both of them when running `ncu-team sync README.md`

PR #45, opened 2018-03-06
doc: add MANIFESTO.md

MylesBorins:
I wanted to take a shot at finding some core principles that our
group can rally behind. Please feel free to push a commit to add
your name to the signatories. Also please comment if anything doesn't
work for you.


MANIFESTO.md
```diff
@@ -0,0 +1,33 @@
...
+  - art and music
+* Getting modules "right" is hard because of these varying interests
+* Creating a good developer experience is the number one priority
+* The aim will be to not compromise the Node.js Developer Experience, but we recognize that may be necessary to improve the health of the JavaScript ecosystem
+* While compromise is neccessary, it is important to recognize when Node.js needs to lead and attempt to influence change at an industry level
+* The closer the Node.js and Web platform remain the better the developer experience will be
+
+* The Node.js implementation of ESModules must have the following features:
+  - Compliant to ecma262. It is important to follow the standard. If we need changes in the standard we should attempt to accomplish them at tc39.
+  - Interoperability. It is important to be able to utilize ESModule code from CommonJS and vice versa.
```
WebReflection: I don't agree about this. ESM being able to import CJS might be fine but CJS able to use ESM is a non sense to me. CJS that needs ESM should upgrade to ESM. This is the only way forward.

I won't start a review because I don't require changes, I just want to point out this is the biggest mistake I can see in this whole story and the reason we are still discussing this whole story.

jdalton:
> The Node.js implementation of ESModules must have the following features:
> * Compliant to ecma262. It is important to follow the standard. If we need changes in the standard we should attempt to accomplish them at tc39.

The current approach of going to standards bodies with tiny asks without broader context or working implementations to back it up hasn't worked out well. I'd like to see that changed to implement first and then go in with asks.

bmeck:
@jdalton thats what the staging system now encourages during stage 3. Is there any specific task you would like to see here? Do we need to change the text in this document somehow to clarify? I don't think we can really make much in terms of breaking spec changes to existing things from my experiences and given that browsers have started shipping ESM unflagged. As long as that is kept represented I think changing text would be fine, but am unsure on what is exactly being asked.

MANIFESTO.md
MylesBorins: we will always have dynamic import as a mechanism to get ESM into CJS. As far as I'm concerned that satisfies this

jdalton:
>  Do we need to change the text in this document somehow to clarify? 

Yes. If we focus on developer experience and scenarios the spec changes will fall out of that naturally. While we should try to stay in the bounds of today's spec we shouldn't be shackled to it. We should have to freedom to implement to fulfill the dev experience/scenario and if spec asks emerge we can take them to the standards bodies with fully realized implementations.

devsnek:
nothing in the current manifesto says we can't do that, all it says is that our final implementation must be spec compliant and that if we want to change stuff we should work to do that

MANIFESTO.md
WebReflection: OK, fair enough, that "_modern2quirk_" browsers do as well, but it's not crystal clear in this point that's all we want to do. Or better, I'd rephrase that:

> It is important to have backward compatibility and utilize CommonJS from ESModule code and, where possible, vice versa.

That'd be a better phrasing to me, otherwise everything else is more than fine to me.

jdalton:
> Compliant to ecma262

Could be

> Compliant to ecma262 as possible

which I think fits the spirit a bit more.

bmeck:
@jdalton I really want to keep strict compliance. We can iterate and implement something that isn't to spec and seek spec changes, but would want to remove anything that isn't able to be compliant before releasing a final implementation. I think implementing something that pushes things outside of the spec or requires backwards compatible changes is fine, but not something we should ship if we cannot change the spec. As long as we agree that the final implementation should be spec complaint, I think going to the committee with a full implementation that is not spec compliant to request changes is fine. We would still need to get browsers on the committee to agree to changes, which has been the problem in the past.

jdalton:
> As long as we agree that the final implementation should be spec complaint, I think going to the committee with a full implementation that is not spec compliant to request changes is fine. 

Node should be able to put its user experiences/scenarios first and should not be roadblocked by other parties. Node should try to the best of its ability to work within existing constraints and with the respective standards bodies but if push comes to shove Node's user experiences/scenarios should win.

robpalme:
> Compliant to ecma262 as possible.

It would be a bit odd to have a manifesto that starts by recognising an existential risk to JavaScript then state that we might need to violate the specification of that language.

jkrems:
> Node's user experiences/scenarios should win.

While I do agree in general (we should focus on trying to solve our user's problems), I don't think it's good to imply that we're matching the spec "just because". We're trying to match the spec because we assume that it *is* in the interest of users that their code is likely to work outside of node and be compatible with future iterations of the language.

jdalton:
I don't anyone stated "just because". I'm aware that following the spec is a good thing for users too but it's not the sole driving force of all the things. I think "as possible" captures that spirit without being overly asterisked and allows Node to fulfill its dev scenarios/experiences to the fullest.

MANIFESTO.md
```diff
@@ -0,0 +1,34 @@
...
+  - Hooks. It is important to be able to have hooks into the lifecycle of a Module.
+  - Multiple Goals. e.g. wasm. It is important that our implementation scales to any number of future goals.
+* Packages written with the Node.js implementation of ESModules must have the following features.
+  - Dual mode. It is important that a single package should be capable of being loaded as both ESM and CJS. This does not have to be the default.
```
ljharb: I'm not sure about "This does not have to be the default"; I think that should be removed. "should be capable", imo, already covers the important part without bogging us down in the implementation detail about defaults.

MylesBorins:
One thing I think is worth mentioning... nothing that is currently considered contentious is due to the spec... if anything it is the things that are not specified that we are having the hardest time reaching consensus on (module specifier resolution).

There is nothing saying that our implementation cannot be a superset and still be compliant. For example, while the tc39 decided against using an in band identifiers such as `use module`, there is nothing stopping core from implementing it, aside from a lack of consensus.

MANIFESTO.md
```diff
@@ -0,0 +1,34 @@
...
+  - Web Compatible. It is important that a published package be capable of running on both the web and node.js platform without any source changes. This does not mean all APIs need to be compatible.
```
ljharb: "Without any source changes" is a great goal; but it's only possible if both browsers and node allow customizable loaders so that someone can have node-like import paths and have that work in the browser via a runtime loader, or have browser-like import URLs and have that work in node via a runtime loader.

Absent loaders, I think "without any source changes other than import paths" would be an acceptable place to land. It would be nice to capture that here.

MylesBorins: the import part here, for me, was the inclusion of "be capable". The goal is not that all node scripts be able to run in the browser, it is that a script can be authored that will simply run in both

MANIFESTO.md
MylesBorins: Let me think on this. I didn't think that "should be capable" was strong enough on its own. Funny enough this is the opposite communication problem of your comment below

MANIFESTO.md
MylesBorins: I'll think on this. Thanks for bringing it up

MANIFESTO.md
ljharb: Sure. Until loaders exist in node and/or browsers tho, "without any source changes other than import paths" is all that's possible, and the path to those things existing might be longer than we all expect.

weswigham: > The goal is not that all node scripts be able to run in the browser, it is that a script can be authored that will simply run in both

Trivially:
`void 0`

You need specific bounds for the goal to be useful.

MANIFESTO.md
weswigham: > CJS that needs ESM should upgrade to ESM. This is the only way forward.

> we will always have dynamic import as a mechanism to get ESM into CJS

Asynchrony is toxic - once you introduce it it has to be everywhere (making the transition a pain). TBQH, if the original spec allowed synchronous resolution, I don't think anyone would be making these kinds of statements, since cross-talk would just be good manners. There's _a_ way, even if it's not a great one for any kind of transitional period, however, which is important.

MANIFESTO.md
weswigham: `Dual mode` seems wanting in the context of `Multiple Goals`. What about my `wasm` implementation that I want to ship alongside my `esm` and `cjs` ones? Why is the limit two?

ljharb: +1, it should be "multi mode".

MANIFESTO.md
benjamn: I agree with @weswigham that it would be nice if there was a way to import ESM synchronously, but that ship seems to have sailed. Forcing Script code to use dynamic `import()` to import ESM seems like a fine interop story, all things considered.

It seems much more important that ESM should be able to use `import` declarations to import CJS—and not only CJS, but a variety of other script/module formats, including WASM and/or anything else the ecosystem dreams up in the future. In other words, we should let go of any assumption that `import` can only be used to import other ESM modules.

In order to succeed as a module format, ESM does not need to be easily importable by non-ESM code, but it absolutely does need to be able to import other popular formats. If CJS makes it significantly easier to import other CJS or WASM code, CJS is going to retain its considerable competitive advantage over ESM for that much longer. If ESM can easily import any other kind of module, the reasons against writing ESM will quickly disappear.

weswigham: > Forcing Script code to use dynamic import() to import ESM seems like a fine interop story, all things considered.

It's not really _fine_ but it's where it is today. The lack of transparent interop is going to seriously hurt the ecosystem, since it means a package deciding to only ship esm is always a breaking change, even on new host versions. 😦

MANIFESTO.md
```diff
@@ -0,0 +1,35 @@
...
+  - command line interfaces
+  - libraries and frameworks
+  - transpilers
+  - robots
+  - art and music
+* Getting modules "right" is hard because of these varying interests
+* Creating a good developer experience is the number one priority
+* The aim will be to not compromise the Node.js Developer Experience, but we recognize that may be necessary to improve the health of the JavaScript ecosystem
+* While compromise is neccessary, it is important to recognize when Node.js needs to lead and attempt to influence change at an industry level
+* The closer the Node.js and Web platform remain the better the developer experience will be
```
mcollina: I think this should be the first point.

MANIFESTO.md
benjamingr: I strictly think it should be possible to include ESM from CJS and vice versa, the "how" might be convoluted but the capability should be there and there should be an obvious (and preferably only one) way to do it.

MANIFESTO.md
bmeck: If you want it to be first, would you want to change :

> Creating a good developer experience is the number one priority

?

MANIFESTO.md
```diff
@@ -0,0 +1,41 @@
...
+
+* The Node.js implementation of ESModules must have the following features:
+  - Compliant to ECMA-262. It is important to follow the standard. If we need changes in the standard we should attempt to accomplish them at TC39.
+  - Interoperability. It is important to be able to utilize ESModule code from CommonJS and vice versa.
+  - Hooks. It is important to be able to have hooks into the lifecycle of a Module.
```
bmeck: I might remove the word `lifecycle` since there might be hooks into the system itself outside of any given module lifecycle events (support for various formats without having a specific module to load comes to mind).

MANIFESTO.md
bmeck: are we trying to specify these in the manifesto? should we leave these to something like the goals document / use cases document?

MANIFESTO.md
mcollina: actually, after that one.

MANIFESTO.md
MylesBorins: can you give me good specific language to replace it with, more than happy to be less specific

bmeck: Maybe: 

Hooks. It is important to be able to instrument the module system with well defined integration points.

?

MANIFESTO.md
MylesBorins: Is it important that it moves to the top? I personally liked the current flow outlining the problem and expanding to developer experience as a focus.

We already have `Creating a good developer experience is the number one priority` three points above this

MANIFESTO.md
MylesBorins: If people are ok I'd like to leave this as is. I don't think getting into specifics or implementation is important for this doc. Further I don't think the idea that you should be able to utilize esm <> cjs is controversial

MANIFESTO.md
MylesBorins: I was attempting to focus on the highest level set of requirements that seemed to be universal / non-controversial.

The TLDR on this one was `node should not knowingly break the web`

I can rephrase if that is more to people's taste

MANIFESTO.md
MylesBorins: what about
` Multi mode. It is important that a single package should be capable of exposing more than one loadable interface such as CJS or ESM. This does not have to be the default.`

FWIW I am a bit concerned that we are getting too deep into implementation on this one.

MANIFESTO.md
benjamingr: I vote for the current phrasing

MANIFESTO.md
```diff
@@ -0,0 +1,41 @@
+## The Module Manifesto
+
+We the undersigned agree on the the following,
+
+* Getting modules "right" is one of the biggest existential threats to JavaScript
+* This affects a large number of people on this planet (and future planets), with varying interests including but not limited to:
```
mhdawson: Could we just make it
This affects a large number of people with varying interests including but not limited to:

ljharb: not bullish on interplanetary exploration?

mhdawson: just not my style :)

MANIFESTO.md
```diff
@@ -0,0 +1,45 @@
...
+* Creating a good developer experience is the number one priority
+* The aim will be to not compromise the Node.js Developer Experience, but we recognize that may be necessary to improve the health of the JavaScript ecosystem
+* While compromise is neccessary, it is important to recognize when Node.js needs to lead and attempt to influence change at an industry level
+* The closer the Node.js and Web platform remain the better the developer experience will be
+
+* The Node.js implementation of ESModules must have the following features:
+  - Compliant to ECMA-262. It is important to follow the standard. If we need changes in the standard we should attempt to accomplish them at TC39.
+  - Interoperability. It is important to be able to utilize ESModule code from CommonJS and vice versa.
+  - Hooks. It is important to be able to have hooks into the lifecycle of a Module.
+  - Multiple Goals. e.g. wasm. It is important that our implementation scales to any number of future goals.
```
chrisdickinson: (I might suggest "does not preclude future goals" vs. requiring that our implementation of modules scales, as-is)

MANIFESTO.md
chrisdickinson: I suspect I disagree on the value of dual mode package loading. (It sounds good in theory, but in practice...?) I think our stated goal should be to be able to provide guidance to authors who wish to support CJS and ESM, not necessarily to subscribe to a solution here.

bmeck: @chrisdickinson even if we can only document a recommendation we should agree on how to do this and see if it requires anything from our implementation. I agree that we will probably continue to disagree on this. At work we have lots of packages that ship multiple modes/forms under a single installation of a package. Is there a rephrasing that could be acceptable while we continue to discuss this?

MANIFESTO.md
chrisdickinson: I think this should be rephrased, since it sort of implies the question "To what end?" — why are we trying to avoid source changes? (We rely on source changes in order to support older browsers, to create bundles, to include Node compatibility shims in the browser, etc. — they're load-bearing, but I'm not sure that's a bad thing?) This might be another place where our goal should be a coherent story, rather than a technical constraint.

MANIFESTO.md
chrisdickinson: "A coherent story around adoption, with specific recommendations for package authors who wish to support ES modules in existing packages."

bmeck: I don't like the word "recommendation" since it seems to imply a lack of support on implementation which is the crux of our disagreement. I am unable to think of a word that adequately would replace it and not imply implementation though.

MylesBorins:
I removed all the implementation details which had the amazing effect of solving all of the above conflict. it also felt like picking a subset was a good way to stay true to the copy people have already signed.

That being said please let me know if you think I should start the signature from scratch

jdalton:
I don't think the signatures are needed. We're already members of the WG.
It seems very cliquey.

MylesBorins:
@jdalton that was definitely not the intent. I simply didn't want to land a document like this without fill buy in from the team. (Not just lazy consensus)

This was an attempted to align everyone, and if I'm missing the mark on that perhaps we should close this

MylesBorins:
I'm going to close this. I think it was a good exercise but unsure that we should land it.

Let me know if I should reopen

devsnek:
if we can't decide what to put in a super abstracted manifesto we certainly won't get any real work done. we should work to finish and merge this imo

robpalme:
I agree with Gus. Surely there is some broad statement of intent that we
all believe in. This document is a useful assertion of that. If new or
existing members disagree with the content and are encouraged to explain
why, that's helpful.

ljharb:
We should definitely reopen it, and merge it.

Being a member of the WG signifies merely that you're interested in advancing, blocking, discussing, etc, use cases and implementations.

Signing a letter of intent signifies you're part of a united front of common principles we can all adhere to.

PR #53, opened 2018-03-28
doc: add meeting notes for 2018-03-28

MylesBorins:


doc/meetings/2018-03-28.md
```diff
@@ -0,0 +1,69 @@
...
+### Discussion
+
+* Membership requirements [#43](https://github.com/nodejs/modules/issues/43)
+  - 10 minute timebox
+  - What should we do to make sure that we are able to maintain quorum
+  - Is Observer sufficient
+  - Punting to next week when more people are around
+* Package-Name-Maps a proposal for bare imports in browsers [#51](https://github.com/nodejs/modules/issues/51)
+  - 10 minute timebox
+  - Presentation from Justin
```
TimothyGu: @justinfagnani Would be great if you could post a link to your talk!

doc/meetings/2018-03-28.md
```diff
@@ -0,0 +1,69 @@
...
+### Approving PRs ~ 1 minute timebox
+
+* doc: add meeting notes for 2018-03-14 [#50](https://github.com/nodejs/modules/pull/50)
+  - make pr to update rules around meeting notes
+
+### For awareness ~ 3 minute timebox per item
+
+#### PRs against Core
+
+* module: fix `e.stack` error when throwing undefined or null [#19282](https://github.com/nodejs/node/pull/19282)
```
ljharb: what was the decision here? I'm still not sure why a PR that should be handling falsiness is removing a side-effecting getter call.

jdalton: We did not have quorum so we used it as a review for the PRs.

robpalme:
At the end, in the discussion of interop and `require()`'s ability to synchronously load ESM, someone (I'm not sure who) stated that Top-Level Await will prevent this. @jdalton responded that @std/esm already solves this, "what can I tell you".

Having checked @std/esm I see it only implements TLA for modules without the `export` keyword.  So it seems like the original point stands.

jdalton:
> Having checked @std/esm I see it only implements TLA for modules without the export keyword. So it seems like the original point stands.

To be clear I support [variant C of the stage-1 proposal](https://github.com/MylesBorins/proposal-top-level-await#variant-c-top-level-await-can-only-be-used-in-modules-without-exports). It seems to do the trick.

robpalme:
Ok - thank you for clarifying.

PR #58, opened 2018-04-03
doc: update members list

MylesBorins:
* sync with with ncu-team sync

> \#31 should have added @thelarkinn to the members list. It was closed
for unrelated reasons but we never got Sean on the list.

* s/Modules Team Members/Members

* remove hyperlink from "Members" header

>The original hyperlink lead to:
>
>https://github.com/orgs/nodejs/teams/modules/members
>
>Which is only accessible to people in the Node.js org afaict. The
member list is in the doc, the link is likely to just confuse folks.

* update ncu-team meta data

>It now can inline Members and Observers. This is connected to the ACL
in github. The Observers header is currently commented out.

* fix link in governance to membership list

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,7 +1,7 @@
 ### Modules Team
 
 For the current list of Team members, see the project
-[README.md](./README.md#current-project-team-members).
+[README.md](./README.md#members).
 
 ### Collaborators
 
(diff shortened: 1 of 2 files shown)


MylesBorins:
fixed a few more errata... initial post is now up to date with all changes

PR #63, opened 2018-04-11
Add Geoffrey as an observer

GeoffreyBooth:
Opening a PR for this feels overly dramatic . . .

PR #65, opened 2018-04-12
Add Sven as an observer

xtuc:
I'm in the Babel and WebAssembly space.

Depends on #63


mcollina:
SGTM, but I think the member list is auto generated.

MylesBorins:
@mcollina I'll update the underlying github permissions... we can use this PR to track approval

MylesBorins:
Landed in README

PR #88, opened 2018-05-19
Add Kevin Smith as an observer.

jdalton:
This adds @zenparsing as an observer.


zenparsing:
Thanks, @jdalton! Yes, I believe that the work being done here is crucial to the future of the language and would like to observe.

PR #128, opened 2018-06-11
governance: s/pull request/issue for new members

MylesBorins:
Our current process for opening a PR to request membership was
implemented before we started using ncu-sync to keep the membership
list up to date. With this is mind I think it makes more sense
to have new members make a request to join via issue rather than
PR.


ljharb:
This has received quorum; thus I believe it can be merged immediately. We can of course also wait until it's mentioned at the meeting today :-)

PR #134, opened 2018-06-18
Add features list to README

### LINKED ISSUE #55: Use Cases (Tracking Issue)
Let's use this issue to track our Use Cases discussion. We will be iterated on the use cases in a [google doc](https://docs.google.com/document/d/10BBsIqdAXB9JR2KUzQGYbCiVugYBnxE4REBakX29yyo/edit) and discussing meta points in this thread.

This will be an ongoing agenda item until we have consensus on core use cases.

Google Doc: https://docs.google.com/document/d/10BBsIqdAXB9JR2KUzQGYbCiVugYBnxE4REBakX29yyo/edit

This trumps https://github.com/nodejs/modules/issues/41 which was originally about making a standalone meeting

### LINKED ISSUE #129: Closing Feature Issues
We currently have, at my count, over 25 open issues for discussion about features. A number of these never developed into a conversation, the remaining conversations have mostly gone stale.

I would like to suggest that we close these issues in order to keep the issue tracker more manageable, and open issues about specific features as we are discussing them

Thoughts?

GeoffreyBooth:
Resolves #129.

Once this is merged in, per #129 we would close all the feature issues. We can continue discussing features on those threads, but there won’t be a need to keep the GitHub issues themselves in an open state.

There are still some use cases that haven’t been discussed in any meeting, because the authors of those use cases weren’t present to explain them; so there will probably be a few more features added to this list in the future. We can also decide to cross off or remove features from this list, but I’d rather this PR not involve that discussion, if people don’t mind? This PR and commit can represent the features list as it exists today, and revisions to this list can get their own PRs and discussions in those future PRs’ threads.

Once this is merged in we can also close #55, as there would be a link to it (and its referenced Google Doc) from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,43 @@ Work includes:
 * Node.js and Browser interoperability
 * VM Modules implementation
 
+## Features
+
+Based on [these use cases](https://docs.google.com/document/d/10BBsIqdAXB9JR2KUzQGYbCiVugYBnxE4REBakX29yyo/edit) ([#55](https://github.com/nodejs/modules/issues/55)), our implementation aims to support the following features:
+
+- Spec compliance ([#132](https://github.com/nodejs/modules/issues/132))
+- Browser equivalence ([#133](https://github.com/nodejs/modules/issues/133))
+- No refactoring ([#87](https://github.com/nodejs/modules/issues/87))
+- Pluggable Loaders to support multiple use cases ([#82](https://github.com/nodejs/modules/issues/82))
+- Named exports when importing CJS ([#81](https://github.com/nodejs/modules/issues/81))
+- Dual-goal packages ([#93](https://github.com/nodejs/modules/issues/93))
+- Multi-mode packages ([#94](https://github.com/nodejs/modules/issues/94))
+- Code coverage/instrumentation ([#95](https://github.com/nodejs/modules/issues/95))
+- Runtime loaders, transpilation at import time ([#96](https://github.com/nodejs/modules/issues/96))
+- Arbitrary sources for module source text ([#97](https://github.com/nodejs/modules/issues/97))
+- Mock modules (injection) ([#98](https://github.com/nodejs/modules/issues/98))
+- Mixed module types within app/module; gradual migration from CommonJS to ESM ([#99](https://github.com/nodejs/modules/issues/99))
+- Transparent interoperability for ESM importing CommonJS ([#100](https://github.com/nodejs/modules/issues/100))
+- Polyfillability ([#101](https://github.com/nodejs/modules/issues/101))
+- Tree shaking ([#102](https://github.com/nodejs/modules/issues/102))
+- File / path / URL resolving ([#103](https://github.com/nodejs/modules/issues/103))
+- Retrievable module metadata ([#104](https://github.com/nodejs/modules/issues/104))
+- Transparent migration ([#105](https://github.com/nodejs/modules/issues/105))
+- WASM modules ([#106](https://github.com/nodejs/modules/issues/106))
+- Browser and Node compatibility without building ([#107](https://github.com/nodejs/modules/issues/107))
+- Browser-compatible build that includes CommonJS ([#108](https://github.com/nodejs/modules/issues/108))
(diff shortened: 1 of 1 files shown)


bmeck:
As long as we understand that the list of features may end up changing due to feasibility, conflicts with other features, etc. might mean crossing out some or adding others, this seems good.

README.md
```diff
@@ -13,6 +13,43 @@ Work includes:
 * Node.js and Browser interoperability
 * VM Modules implementation
 
+## Features
+
+Based on [these use cases](https://docs.google.com/document/d/10BBsIqdAXB9JR2KUzQGYbCiVugYBnxE4REBakX29yyo/edit) ([#55](https://github.com/nodejs/modules/issues/55)), our implementation aims to support the following features:
```
MylesBorins: To brad's point I think it is important that we are explicit that this list is subject to change.

I also see features in here that I do not believe we have consensus on, the statement at the top seems to imply that we are aiming to support all the below features... that seems inconsistent

README.md
```diff
@@ -13,6 +13,43 @@ Work includes:
...
+- Mixed module types within app/module; gradual migration from CommonJS to ESM ([#99](https://github.com/nodejs/modules/issues/99))
+- Transparent interoperability for ESM importing CommonJS ([#100](https://github.com/nodejs/modules/issues/100))
+- Polyfillability ([#101](https://github.com/nodejs/modules/issues/101))
+- Tree shaking ([#102](https://github.com/nodejs/modules/issues/102))
+- File / path / URL resolving ([#103](https://github.com/nodejs/modules/issues/103))
+- Retrievable module metadata ([#104](https://github.com/nodejs/modules/issues/104))
+- Transparent migration ([#105](https://github.com/nodejs/modules/issues/105))
+- WASM modules ([#106](https://github.com/nodejs/modules/issues/106))
+- Browser and Node compatibility without building ([#107](https://github.com/nodejs/modules/issues/107))
+- Browser-compatible build that includes CommonJS ([#108](https://github.com/nodejs/modules/issues/108))
```
MylesBorins: Do we have consensus on this feature?

README.md
```diff
@@ -13,6 +13,43 @@ Work includes:
...
+- Browser-compatible specifier resolution ([#109](https://github.com/nodejs/modules/issues/109))
+- Specifier resolution customization ([#110](https://github.com/nodejs/modules/issues/110))
+- Package encapsulation ([#111](https://github.com/nodejs/modules/issues/111))
+- Don’t break CommonJS ([#112](https://github.com/nodejs/modules/issues/112))
+- Conditional imports ([#113](https://github.com/nodejs/modules/issues/113))
+- Import JSON without needing asynchronous syntax ([#114](https://github.com/nodejs/modules/issues/114))
+- Importing non-JavaScript files ([#115](https://github.com/nodejs/modules/issues/115))
+- Import CommonJS without needing asynchronous syntax ([#116](https://github.com/nodejs/modules/issues/116))
+- NodeJS contextual pathing use cases ([#121](https://github.com/nodejs/modules/issues/121))
+- Load ESM over https ([#127](https://github.com/nodejs/modules/issues/127))
```
MylesBorins: Do we have consensus on this feature?

xtuc: Can we discuss later today about this? if it's actually a feature or not

GeoffreyBooth:
@MylesBorins I added a note that the list is subject to change. I don’t think there should be language that the features lack consensus, or that the final implementation may not achieve all of them. Since this list _is_ subject to change, I think features that lack consensus should be removed after discussion. When we get to the point of a final implementation, if it doesn’t support some of these features we should again discuss them, and decide if those features _should_ be supported (and therefore we keep working on the implementation) or if at that point they should be dropped.

As for the particular features that you pointed out, can we please merge this in first so that I can close all the feature issues? And then you can open another PR to suggest removing some features, and we can discuss there.

MylesBorins:
@GeoffreyBooth I would prefer we remove features that do not have consensus and open prs specifically to add them.

The `Load ESM over https` in particular is the one I would like to see removed

GeoffreyBooth:
@MylesBorins Updated.

PR #158, opened 2018-07-25
Create Terminology.md

SMotaal:
Based on last week's meeting, and based on some testing, this PR would add `doc/Terminology.md` to the main repo. Please note that we will have to use inline HTML tags for since GFM does not provide respective syntaxes for `<dl>`, `<dt>` and `<dd>`. Both AsciiDoc and RST formats were considered since unlike GFM they actually have syntaxes for definitions, but they bring their own complexities and limitations that made them much less ideal for this document.

_Edit (Fishrock123): [**Rendered Markdown**](https://github.com/nodejs/modules/blob/terminology/doc/Terminology.md)_

### DIFF
--- a/doc/Terminology.md
+++ b/doc/Terminology.md
@@ -55,14 +55,18 @@ Normative Terms
   Consumer-defined Disambiguation
 </dt>
 <dd>
+  
   When importing a file into an ESM context, the parse goal of the file is determined by the importing module (the consumer). This is often accomplished via syntax, for example by using only `import` statements for ESM or `require` (or other function) for CommonJS.
+  
 </dd>
 
 <dt id="author-disambiguation">
   Author-defined Disambiguation
 </dt>
 <dd>
+  
   A file or module’s parse goal is included within itself. This is often achieved via a file extension such as `.mjs` or a field in a `package.json` file. Other suggestions have included a `"use module"` directive or unambiguous syntax, such as parsing the file for `import` or `export` statements.
+
 </dd>
 </dl>
   
--- a/doc/Terminology.md
+++ b/doc/Terminology.md
@@ -56,13 +56,14 @@ Normative Terms
 </dt>
 <dd>
   When importing a file into an ESM context, the parse goal of the file is determined by the importing module (the consumer). This is often accomplished via syntax, for example by using only `import` statements for ESM or `require` (or other function) for CommonJS.
-</dl>
+</dd>
 
 <dt id="author-disambiguation">
   Author-defined Disambiguation
 </dt>
 <dd>
   A file or module’s parse goal is included within itself. This is often achieved via a file extension such as `.mjs` or a field in a `package.json` file. Other suggestions have included a `"use module"` directive or unambiguous syntax, such as parsing the file for `import` or `export` statements.
+</dd>
 </dl>
   
 Deprecated Terms
--- a/doc/Terminology.md
+++ b/doc/Terminology.md
@@ -1,0 +1,64 @@
+Terminology
+============
+
+This document is used to keep track of common terminology related to modules in [Node.js](https://nodejs.org), focusing primarily on ECMAScript Modules as well as other mainstream module formats.
+
+<dl>
+
+Normative Terms
+---------------
+  
+<dt id=agnostic-consumers>
+  Agnostic Consumers
+<dd> 
+
+  The ability for importer modules to be unaware of the imported module type (CJS vs ESM). This implies that the imported module can be migrated from CJS to ESM without impacting the consumer.
+
+  This term alone does not specify in which direction(s) the agnosticism applies.
+
+<dt id=require-interop>
+  Require Interoperability `require(<esm>)`
+<dd>
+  
+  The ability for a module to import an ESM module using the `require(…)` function.
+
+  The module’s namespace is return directly, i.e. it is not wrapped in a promise.
+
+  If it is impossible to load the graph synchronously, it is intended that the call must throw.
+
+<dt id=import-interop>
+  Import Interoperability `import(<cjs>)`
+<dd>
+  
+  The ability for an ESM module to import a CJS module.
+
+  Applies to both static `import … from ` and dynamic `import(…)`.
+
+  This term alone does not specify whether named `exports` are extracted from the CJS exports object.
+
+<dt id="named-exports">
+  Named Exports `exportKeys`
+<dd>
+  
+  The ability for a CJS module to elect to expose properties of its `exports` object as fixed set of named exports for use by ESM consumers, as opposed to having a single `export default` equal to the exports object.
+
+  Some forms of CJS are not amenable, e.g. those that dynamically delete properties of the exports object.
+  
+</dl>
+
+<dl>
+  
+Deprecated Terms
+----------------
+
+<dt id=transparent-interop>
+  Transparent Interoperability
+<dd>
+        
+  An imprecise term that refers to some subset of the following independent capabilities:
+  
+  1. [Agnostic Consumers](#agnostic-consumers) (in one or both directions)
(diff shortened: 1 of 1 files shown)


doc/Terminology.md
```diff
@@ -0,0 +1,64 @@
...
+
+  The ability for importer modules to be unaware of the imported module type (CJS vs ESM). This implies that the imported module can be migrated from CJS to ESM without impacting the consumer.
+
+  This term alone does not specify in which direction(s) the agnosticism applies.
+
+<dt id=require-interop>
+  Require Interoperability `require(<esm>)`
+<dd>
+  
+  The ability for a module to import an ESM module using the `require(…)` function.
```
GeoffreyBooth: I thought `require` always imported CommonJS.

ljharb: `require` brings in whatever's in `require.extensions` - that's how most coffeescript users bring in their coffeescript, i believe. By default, it can bring in CJS, or JSON, or `.node` (binary modules). Adding ESM to that list would achieve "require interop" by default.

guybedford: Yes we need a meaningful term to refer to this, even if it is not implemented.

I'd suggest switching around the order here and including `Import Interoperability` first.

Perhaps it's even worth a note here that this has async concerns for implementation and linking to some reference?

GeoffreyBooth: I think it’s worth somewhere explaining that there are two general approaches to `import` and `require`. The approach of `--experimental-modules` plus `import.meta.require`/`createRequireFunction` is that in an ESM context, `import` can bring in anything—more ESM, CommonJS, etc.—and `require` can bring in at least CommonJS and maybe also ESM. Hence the need for file extensions to disambiguate what’s being imported.

The approach of the NPM implementation and others is that `import` can only bring in ESM, and `require` (whether via `import.meta.require` or `createRequireFunction` or something else) can only bring in CommonJS. The idea here is that the method doing the importing disambiguates the parse goal, and therefore file extensions are irrelevant.

People have very strong opinions about either of the above, which I think we can keep out of a neutral terminology document. Both approaches are common enough across the various proposals and implementations that they should each be covered here.

ljharb: There’s a need for many tools to know the parse goal of a stand-alone file - absen any of its consumers - so file extensions would always be relevant.

Fishrock123:
I have added this to the OP:

> _Edit (Fishrock123): [**Rendered Markdown**](https://github.com/nodejs/modules/blob/terminology/doc/Terminology.md)_

doc/Terminology.md
GeoffreyBooth: > There’s a need for many tools to know the parse goal of a standalone file - absent any of its consumers - so file extensions would always be relevant.

I think this is the subtext in the discussion over `.mjs`. There are some in this group who see `.mjs` as a feature in and of itself, because of this file-level parse goal disambiguation that it offers; and others who view it as an implementation detail, as just one way of several for the runtime to determine a file’s parse goal. For “many tools,” sure, it might be nice to be able to determine the parse goal from a file extension. But that’s a separate goal, outside the scope of bringing ES modules support to Node.

guybedford: The goal of this document is to try and provide a glossary of technical terms that get used commonly in these discussions to ensure they have well-defined and known meanings, as opposed to trying to describe the entire landscape of proposals. If you want to suggest new terms to add here, please do - that's what it is for.

GeoffreyBooth: @guybedford Yes, and I’m pointing out that we shouldn’t define the terms such that they reflect only one way of thinking. There are obviously other points of view on this matter. Babel and eslint currently manage `import` and `export` statements without needing `.mjs`, and they can continue to do so. The NPM and other implementations make do without `.mjs`. It’s simply not the case that this is required.

ljharb: @GeoffreyBooth they do so with external information in their configuration, which is not acceptable for standard node support of modules. It's absolutely required.

guybedford: @GeoffreyBooth I get what you're saying, and while I disagree that terms need to be represented equally between proposals (some proposals naturally have more terms, some don't have any), it would help if you could explicitly suggest specific terms from other proposals or side notes to this effect rather than asking someone else to do it. But explaining proposals or what the layout of proposals looks like is not a goal here.

GeoffreyBooth: Is there a proposal out there where `require` can be used to import ESM? I’m not familiar with one. All the proposals I’m aware of that involve using `require` in an ESM context (NPM, @MylesBorins, etc.) use it to import CommonJS only, not either CommonJS or ESM. I’m happy to submit a PR to update the document but I don’t want to do so if I’m incorrect.

ljharb: Yes, require should be able to bring in ESM also, ideally.

GeoffreyBooth: > Yes, require should be able to bring in ESM also, ideally.

If no one’s ever proposed it, I don’t see why it would be in our glossary. I’ve only ever seen `require` in an ESM context discussed as a way of bringing in CommonJS only, usually as a way of keeping CommonJS out of `import` statements.

doc/Terminology.md
```diff
@@ -0,0 +1,88 @@
...
+<dd>
+  
+  The ability for an ESM module to import a CJS module.
+
+  Applies to both static `import … from ` and dynamic `import(…)`.
+
+  This term alone does not specify whether named `exports` are extracted from the CJS exports object.
+</dd>
+
+<dt id="named-exports">
```
robpalme: I think we're specifically talking about CJS exposing named exports.  So the generic term ***named exports*** seems too general.  How about renaming the term to ***Named Exports from CJS***?

doc/Terminology.md
```diff
@@ -0,0 +1,133 @@
...
+-----------------
+
+<dt id="browser-interop">
+  Browser Interop
+</dt>
+<dd>
+
+  The expectation that a module (namely ES modules) can be written once and be used in both Node.js and a "popular browser" that is "reasonably conforming", beyond the implicit requirements of conformance to the behaviours defined by the ECMAScript specificion.
+
+  <dl>
```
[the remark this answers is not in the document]
SMotaal: @guybedford @myles what are your thoughts on adding this section, can you recommend a better way for what I think will reflect "contractual" concepts for this team, in the sense that if they make it through PR, they become an obligation to be honoured (unless revised).

Stipulations work here in the sense that we are defining more concrete aspects to otherwise generalized terms in some specific context.

The example I include is "Browser Interop" (with stipulations) not "browser interoperability" (in general) which provides a concrete set of rules, expectations... etc. for the latter when used elsewhere in documents and conversations.

doc/Terminology.md
```diff
@@ -0,0 +1,133 @@
...
+
+Stipulative Terms
+-----------------
+
+<dt id="browser-interop">
+  Browser Interop
+</dt>
+<dd>
+
+  The expectation that a module (namely ES modules) can be written once and be used in both Node.js and a "popular browser" that is "reasonably conforming", beyond the implicit requirements of conformance to the behaviours defined by the ECMAScript specificion.
```
SMotaal: FIXME: typos

jkrems:
Should we land this to allow future additions as PRs? It might also help with visibility. Or are we looking to pushing to this shared branch?

SMotaal:
I would love that, so why don't we get approvals so we can merge by next meeting

SMotaal:
@MylesBorins do we merge this?

MylesBorins:
@SMotaal we cannot merge until it has been reviewed and approved in a meeting with quorum. That is the current governance rules

SMotaal:
That's was my thought, so consider this my virtually **re**adding this to modules-agenda for next meeting 😄

Issue #151, opened 2018-07-12
Feature: ESM in .js files


GeoffreyBooth:
There needs to be a way to tell Node to treat `.js` files as ESM JavaScript.

Besides the issue raised in [#149](https://github.com/nodejs/modules/issues/149), there’s the use case of legacy build tools and pipelines that expect only a single file extension for JavaScript. I can give a concrete example for the case of CoffeeScript. The problem is that essentially, the CoffeeScript compiler is a string converter. It takes a string as input, like `console.log 'hello!'`, and returns another string as output: `console.log('hello!');` People who use the CoffeeScript compiler as their entire build chain (as opposed to using it within Gulp or Webpack or the like) use it to transpile projects like this:

```bash
coffee --compile --output dist/ src/
```

This command recursively searches `src/` for all `*.coffee`, `*.litcoffee` and `*.litcoffee.md` files, runs them through the CoffeeScript compiler, and saves them as `*.js` files in a mirrored folder tree under `dist/`.

CoffeeScript doesn’t know the parse goals of the files in `src/`. It doesn’t know which ones to save with a `.js` extension versus an `.mjs` extension. About the only way it _could_ know is if we created new file extensions, like `.mcoffee`, `.mlitcoffee` and `.mlitcoffee.md`, that it knew to output as `.mjs`. But that’s not a practical solution, as there are thousands of CoffeeScript build plugins and syntax highlighters and so on out there that would need to be updated to support these new file extensions; and let’s be honest, most of those packages are abandoned. Adding a new option, like `--output-file-extension mjs`, also isn’t realistic for the same reason, because of the sheer number of build plugins that would need to be updated to support it. CoffeeScript traditionally just converts one string into another, without necessarily reading or passing along any other options or metadata, and there are lots of build pipelines that expect this behavior.

Rather than figuring out a way to add a piece of metadata (the parse goal) to CoffeeScript source files, and then updating all other parts of the build chain to preserve and pass along and understand this new metadata, it would be better if the parse goal could be specified outside of the file itself.

[...]

PR #160, opened 2018-07-28
Proposals for configuring parse goals of files in --experimental-modules

GeoffreyBooth:
Following up from #150, this thread is for determining the user experience of an addition to `package.json` that configures Node to treat `.js` files as ESM within a package boundary (i.e. with the “module” parse goal). This assumes the `--experimental-modules` implementation as a baseline; other implementations might not have this issue, for example if they don’t allow importing ESM via `import` statements. For suggestions of other implementations, can I invite people to open their own threads for alternate proposals? And let’s keep this one focused on what ESM in `.js` in `--experimental-modules` should look like.

Readable view: https://github.com/GeoffreyBooth/modules/blob/esm-in-js-ux/proposals/esm-in-js-ux.md

To discuss particular proposals, please add comments as PR notes after those proposals. Add general comments to suggest new proposals or broader changes.

___

**Update**: we have reached consensus! Of the people who have commented on this thread, at least. Here’s the consensus proposal:

### MIME types/webserver as metaphor, defined in `package.json` and/or in linked JSON files

In browsers, users must configure their webservers to serve `.js` or `.mjs` files with a JavaScript MIME type like `text/javascript` in order for the browser to recognize the file as ESM. The browsers pay no attention to file extensions, but most webservers use file extensions to determine what MIME types to use to serve files. This is similar to `AddType video/webm .webm` that you may have seen in Apache webserver configuration files.

The idea here is to mimic this webserver configuration in a new `package.json` section called `mimes`:

```json
"mimes": {
  "js": "application/node",
  "mjs": "text/javascript",
  "json": "application/json"
}
```

In this example, `.js` files would be treated as CommonJS, which has the MIME type `application/node`. `.mjs` files would be treated as ESM (`text/javascript`, or `application/javascript`) and `.json` files would be treated as JSON. The configuration in this example would be the default, what Node uses if a `mimes` field was missing, as historically `.js` files were always treated as CommonJS.

Here’s another example:

```json
"mimes": {
  "js": "text/javascript",
  "cjs": "application/node"
}
```

This tells Node to treat `.js` files within this package boundary as ESM, and a new `.cjs` file extension as CommonJS. The `.cjs` extension could be anything—just as in Apache a user could add the configuration `AddType video/webm .foo`, this `mimes` block provides the flexibility for new file extensions to be defined and mapped to MIME types that Node understands. Since `.mjs` isn’t listed here, Node falls back to the default mapping for it (`text/javascript`), and likewise for any other file extensions that Node recognizes by default (`.json`, etc.).

That `mimes` section could alternatively take an array of strings, which would be relative or resolved paths to JSON files:

```json
"mimes": [
  "./my-mimes.json",
  "typescript/mimes.json"
]
```

Each JSON file would be an object with “extension: mime” mappings like the first example. They would be combined using `Object.assign`. If the first element in the array is `null`, Node’s default MIME mappings are discarded first before the array elements are merged, e.g. something like:

```js
const nodeDefaultMimeMappings = require('module').mimeMappings;

const packageJsonMimes = require('./package.json').mimes;
if (Array.isArray(packageJsonMimes)) {
  let mimeMappings = (packageJsonMimes[0] === null) ? Object.create(null) : nodeDefaultMimeMappings;
  packageJsonMimes.forEach((filePath, index) => {
    if (filePath === null && index !== 0) {
      throw new Error('Only the first element of a mimes array may be null');
    }
    Object.assign(mimeMappings, require(filePath));
  });
  return mimeMappings;
} else {
  return Object.assign(nodeDefaultMimeMappings, packageJsonMimes);
}
```

Through this `mimes` section that can take either an object or an array of strings that reference JSON files (of `mimes` mapping objects), users can configure Node’s handling of file extensions.

### DIFF
--- a/proposals/esm-in-js-ux.md
+++ b/proposals/esm-in-js-ux.md
@@ -1,0 +1,75 @@
+These are proposals for the user experience of an addition to `package.json` that configures Node to treat `.js` files as ESM within a package boundary. This assumes the `--experimental-modules` implementation as a baseline. This is spun off from [#150](https://github.com/nodejs/modules/issues/150), and relates to [nodejs/node/pull/18392](https://github.com/nodejs/node/pull/18392).
+
+### [nodejs/node/pull/18392](https://github.com/nodejs/node/pull/18392)
+
+This pull request adds support for a `"mode": "esm"` flag to be added to `package.json`, that tells Node to treat all `.js` files within the package boundary of that `package.json` file as ESM. The package boundary of a `package.json` file is deemed to be all modules in that folder and subfolders until the next nested `package.json` file.
+
+In [#150](https://github.com/nodejs/modules/issues/150) we found consensus on adding a little more customizability than what would be offered by a simple boolean flag, but I wanted to include this here for completeness. Also, I expect that this PR would serve as the basis for a more full-featured implementation.
+
+### MIME types/webserver as metaphor
+
+In browsers, users must configure their webservers to serve `.js` or `.mjs` files with a JavaScript MIME type like `text/javascript` in order for the browser to recognize the file as ESM. The browsers pay no attention to file extensions, but most webservers use file extensions to determine what MIME types to use to serve files. This is similar to `AddType video/webm .webm` that you may have seen in Apache webserver configuration files.
+
+The idea here is to mimic this webserver configuration in a new `package.json` section called `mimes`:
+
+```json
+"mimes": {
+  "js": "application/node",
+  "mjs": "text/javascript",
+  "json": "application/json"
+}
+```
+
+In this example, `.js` files would be treated as CommonJS, which has the MIME type `application/node`. `.mjs` files would be treated as ESM (`text/javascript`, or `application/javascript`) and `.json` files would be treated as JSON. The configuration in this example would be the default, what Node uses if a `mimes` field was missing, as historically `.js` files were always treated as CommonJS.
+
+Here’s another example:
+
+```json
+"mimes": {
+  "js": "text/javascript",
+  "cjs": "application/node"
+}
+```
+
+This tells Node to treat `.js` files within this package boundary as ESM, and a new `.cjs` file extension as CommonJS. The `.cjs` extension could be anything—just as in Apache a user could add the configuration `AddType video/webm .foo`, this `mimes` block provides the flexibility for new file extensions to be defined and mapped to MIME types that Node understands. Since `.mjs` isn’t listed here, Node falls back to the default mapping for it (`text/javascript`), and likewise for any other file extensions that Node recognizes by default (`.json`, etc.).
+
+### Preconfigured MIMEs databases
+
+This is based on [this proposal](https://gist.github.com/bmeck/7ee7eb2147e2dafe3167c856d9b4151a) by @bmeck, and is similar to the previous proposal. In this version, the `mimes` field takes a string or array of strings, where each string indicates a preset MIME database:
+
+```json
+"mimes": "cjs"
+```
+
+This would be the default, which corresponds to the first `mimes` block in the previous proposal (and to the `--experimental-modules` current behavior).
+
+```json
+"mimes": [null, "esm"]
+```
+
+This would blank out any default extension/MIME mapping within the current package boundary, and then apply the `"esm"` preset (which maps both `.mjs` and `.js` to ESM).
+
+### Map from extension to extension
+
+This is based on [this comment](https://github.com/nodejs/modules/pull/150#issuecomment-406838613) from @ljharb. It would map file extensions to other file extensions (@ljharb please expand upon your idea and I’ll update this section, as there wasn’t an example in the comment):
+
+```json
+"extensions": {
+  "js": "mjs"
+}
+```
+
+This would tell Node to treat all `.js` files as `.mjs` files within the package boundary—in other words, to treat all `.js` files as ESM.
+
+### Map from extension to parse goal
+
+Similar to the previous, this would define parse goals for extensions directly:
+
+```json
+"parseGoals": {
+  "mjs": "module",
+  "js": "commonjs"
+}
+```
+
+I presume that this would be limited to JavaScript-like files, unless things like `.json` or `.node` could have parse goals.
--- a/proposals/esm-in-js-ux.md
+++ b/proposals/esm-in-js-ux.md
@@ -73,3 +73,44 @@ Similar to the previous, this would define parse goals for extensions directly:
 ```
 
 I presume that this would be limited to JavaScript-like files, unless things like `.json` or `.node` could have parse goals.
+
+### Consensus: “MIME types/webserver as metaphor” with links to external JSON files
+
+Start with the second proposal from this list, “MIME types/webserver as metaphor”, and its new `mimes` section in `package.json` that can be an object, e.g.:
+
+```json
+"mimes": {
+  "js": "text/javascript",
+  "cjs": "application/node"
+}
+```
+
+That `mimes` section could alternatively take an array of strings, which would be relative or resolved paths to JSON files:
+
+```json
+"mimes": [
+  "./my-mimes.json",
+  "typescript/mimes.json"
+]
+```
+
+Each JSON file would be an object with “extension: mime” mappings like the first example. They would be combined using `Object.assign`. If the first element in the array is `null`, Node’s default MIME mappings are discarded first before the array elements are merged, e.g. something like:
+
+```js
+const nodeDefaultMimeMappings = require('module').mimeMappings;
+
+const packageJsonMimes = require('./package.json').mimes;
+if (Array.isArray(packageJsonMimes)) {
+  let mimeMappings = (packageJsonMimes[0] === null) ? Object.create(null) : nodeDefaultMimeMappings;
+  packageJsonMimes.forEach((filePath, index) => {
+    if (filePath === null && index !== 0) {
+      throw new Error('Only the first element of a mimes array may be null');
+    }
+    Object.assign(mimeMappings, require(filePath));
+  });
+  return mimeMappings;
+} else {
+  return Object.assign(nodeDefaultMimeMappings, packageJsonMimes);
+}
+```
+


proposals/esm-in-js-ux.md
```diff
@@ -0,0 +1,75 @@
...
+
+In [#150](https://github.com/nodejs/modules/issues/150) we found consensus on adding a little more customizability than what would be offered by a simple boolean flag, but I wanted to include this here for completeness. Also, I expect that this PR would serve as the basis for a more full-featured implementation.
+
+### MIME types/webserver as metaphor
+
+In browsers, users must configure their webservers to serve `.js` or `.mjs` files with a JavaScript MIME type like `text/javascript` in order for the browser to recognize the file as ESM. The browsers pay no attention to file extensions, but most webservers use file extensions to determine what MIME types to use to serve files. This is similar to `AddType video/webm .webm` that you may have seen in Apache webserver configuration files.
+
+The idea here is to mimic this webserver configuration in a new `package.json` section called `mimes`:
```
ljharb: Is there a reason not to keep it simple and make it a map of extension to extension?

In other words, js: mjs would load JS dies as if they were mjs files.

devsnek: I'm a fan of mimes because they refer to the type of file, not the extension used, which is already the point of configuration and ambiguity.

proposals/esm-in-js-ux.md
```diff
@@ -0,0 +1,75 @@
...
+"extensions": {
+  "js": "mjs"
+}
+```
+
+This would tell Node to treat all `.js` files as `.mjs` files within the package boundary—in other words, to treat all `.js` files as ESM.
+
+### Map from extension to parse goal
+
+Similar to the previous, this would define parse goals for extensions directly:
```
ljharb: this works except that then someone will have to go look up what the list of parse goals is - instead of representing "i know node wants me to use a .X, but i want to use a .Y instead"

proposals/esm-in-js-ux.md
```diff
@@ -0,0 +1,75 @@
...
+
+This is based on [this comment](https://github.com/nodejs/modules/pull/150#issuecomment-406838613) from @ljharb. It would map file extensions to other file extensions (@ljharb please expand upon your idea and I’ll update this section, as there wasn’t an example in the comment):
+
+```json
+"extensions": {
+  "js": "mjs"
+}
+```
+
+This would tell Node to treat all `.js` files as `.mjs` files within the package boundary—in other words, to treat all `.js` files as ESM.
```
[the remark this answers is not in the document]
GeoffreyBooth: I find this one confusing. Since what this configuration block is doing is _changing_ the parse goal of an extension, I find it confusing to define the intended parse goal _by_ an extension. I don’t think we should use extensions on both sides of the mapping.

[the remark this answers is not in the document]
ljharb: That's a fair counter argument. To me the model would be "i want extension X to pretend it's extension Y" - since without this feature, they'd have to *rename* extension X to extension Y to get identical behavior.

proposals/esm-in-js-ux.md
GeoffreyBooth: Yes, and what _is_ the parse goal of something like JSON? I feel like this one only really covers the cases of `.js` and `.mjs`, in which case it’s not really any better than `"mode": "esm"`. I wanted to include it since it seemed like an obvious option, like someone would presumably ask “why are you suggesting MIME types or file extensions as metaphors for parse goals when you could just ask the user what parse goal they want?” but unless other file types have parse goals, this doesn’t make much sense.

ljharb: It’s “json” - I’m not sure what you’re asking.

This also applies to future things like wasm or html. Every textually formatted file type has a parse goal.

bmeck: @ljharb 
I don't really understand the difference here vs using other things like "mode". Is this argument against all forms of configuring parse goals?

proposals/esm-in-js-ux.md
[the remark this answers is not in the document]
bmeck: @ljharb how is this better than MIMEs? MIMEs are the underlying data and you still need to know them, even if you don't call them by that name. The mental model of ` "i want extension X to pretend it's extension Y" ` still requires people to understand the format of `extension Y` which has some value such as a MIME. Also, I doubt the mental model here is helped by abstracting over reality.

proposals/esm-in-js-ux.md
bmeck: @ljharb you still need to know the format of file extensions, so you still need to keep that data in your head. Also, you need to enable unknown formats. I'm not sure how to enable `.tsx -> .ts` mapping unless loaders only see file extensions instead of MIMEs. Loaders can't always rely on file extensions since some things don't have file extensions, and other situations like `data:` URLs that only have a MIME.

proposals/esm-in-js-ux.md
```diff
@@ -0,0 +1,75 @@
...
+"mimes": "cjs"
+```
+
+This would be the default, which corresponds to the first `mimes` block in the previous proposal (and to the `--experimental-modules` current behavior).
+
+```json
+"mimes": [null, "esm"]
+```
+
+This would blank out any default extension/MIME mapping within the current package boundary, and then apply the `"esm"` preset (which maps both `.mjs` and `.js` to ESM).
```
GeoffreyBooth: As it seems like we’re applying process of elimination, I think this option is less appealing than the one above it (where file extensions are mapped to MIMEs directly, rather than through a database). The direct mapping allows for configuring of new/unknown file types, and is more flexible in general.

proposals/esm-in-js-ux.md
GeoffreyBooth: It seems to me like this is the best of the options on this list: it’s flexible enough to cover not just the ESM-in-`.js` use case, but future unknown types; and even the CommonJS-in-`.cjs` use case, for people who want `.js` to be ESM but still need side-by-side file extensions where one file extension is CommonJS. Most JavaScript developers should be familiar with the concept of defining MIME types for file extensions, as it’s commonly done with webservers, and part of working with files/URLs in browsers.

bmeck: I think this approach has a fairly large weakness in that you can't pull in or delegate your MIMEs to another package, for example if you mix `.coffee` and `.ts` you need to explicitly list each of those rather than pulling in the data from the coffeescript or typescript packages; this delegation would allow mapping of multiple types also in the case of `.coffee` and `.litcoffee` under a single delegation.

Another complaint I have about this approach is that it doesn't expose the delegation mechanism it is relying on for default formats of file extensions. This means that you can't remove file extensions for w/e reasons from resolving, it should have a means of marking an extension as not having a format/unable to resolve such as disabling `.node` files.

proposals/esm-in-js-ux.md
bmeck: How does this not allow for configuring unknown formats / is more flexible? I'm a bit confused at that conclusion. 

The `"mimes": []` allows delegation and bundling of formats into a single string rather than requiring duplication.

In addition it just points to a generic database of formats which could be just like the option above:

```json
// database resolved by `mimes: []`
{
  ".ts": "application/vnd.typescript",
  ".tsx": "application/vnd.typescript;jsx"
}
```

GeoffreyBooth: Aren’t these databases supplied by Node? Can a user specify their own database somehow?

bmeck: Users should be able to configure these, so they should be able to be supplied by configuration outside of Node itself.

GeoffreyBooth: Well then this can also work, but the ability of the user to create their own presets needs to be explained in the proposal.

I feel like most users will only need to remap one or two extensions, not dozens. We could provide both abilities, but I think if the more common use case is just one or two extensions, we should have the other `mimes` proposal and only add this one if there’s a need for presets.

proposals/esm-in-js-ux.md
bmeck: > This would just be increased functionality, wouldn’t it? Like a “MIME preset.” We could have a mimes field as described here, and additionally a mimesPresets field that follows the MIME databases proposal. The individual MIME mappings in the mimes field would override the groups of mappings set by the mimesPresets field.

This is a design difference and isn't suitable for being a purely additive feature: in the approach that uses an object, it is not designed for composition and thus doesn't have a way to delegate. Adding a method to delegate doesn't make sense with the object design and would need to devolve into the `"mimes": []` approach as a separate field:

```json
{
  "mimes": {".jsx": "text/javascript"},
  "mimeDBs": ["typescript"]
}
```

> As far as delegating to other packages, though, why would a user need to list e.g. .coffee in the mimes list? If a CoffeeScript loader were registered, it would handle that extension, including registering the default MIME type that it should have. A user would only need to list .coffee in mimes if they wanted to override the default MIME type that the CoffeeScript loader specifies for .coffee files.

Loaders don't register formats, they just handle requests of `(specifier, referrer) => (format, body)` (based upon [HostResolveImportedModule](https://tc39.github.io/ecma262/#sec-hostresolveimportedmodule) ) and delegate to the parent loader when encountering things, leading to a single authority of formats which is the "default" loader. There isn't a way for a loader to pass `format` to their parent.

> This could be as simple as:

> "mimes": {
>  "node": false
> }

> I’m not sure what use case this serves. Also, couldn’t one create a loader that takes control of .node and chooses to do nothing with it?

The use case it serves is ensure that some format is not used within their packages, things like C++ addons can pose a security risk.

A loader doesn't control a file extension so I'm not sure I understand the second question.

proposals/esm-in-js-ux.md
bmeck: I disagree. Allowing it to be either an object or array works, but adds mental burden. In addition, if we encourage manual configuration rather than delegating to a package, that means that any changes would need to be done in each package rather than in a single location. If you have a JSX format for `.jsx` and wish to disable JSX across all packages you could update all those packages or update a single package containing the mapping. I'm don't think there is the advantage you describe since most people wishing to add 1-2 extensions also have to face updates and generally don't want to know the formats generally, they just want to make it match "coffeescript" or "typescript" etc.

weswigham: > Users should be able to configure these, so they should be able to be supplied by configuration outside of Node itself.

I don't think mime dbs on windows are terribly common - any application which needs one pulls one from somewhere itself, usually (which, for js users usually means the `mime-db` package) - not some global system store of them.

I'm not saying that's bad or good - but that does mean you're effectively just deferring to an npm package to determine what the configuration is, so that should _probably_ be acknowledged in a proposal using mimes.

bmeck: @weswigham I don't think Node/users would generally want a full MIME db like `mime-db`, I would think Node's default DB would just be the supported formats that a Loader can pass to Node and have it properly make a Module Record for. Similarly I don't think they would use a system store for any custom MIMEs. I'm not sure it has to be an `npm` package, having it point to a file on disk would also work, or maybe even a data URL???. This is a bit unclear, but seems a little more specific than the more general discussion going on in this PR.

weswigham: @bmeck Those kind of decisions directly impact the usability of MIMEs and weigh heavily on both implementation and ergonomics. If this PR is discussing implementation options, then those options are equally important, and shouldn't simply be brushed off. If an extension map vs a mode flag are separate options, then so, too, is are a variety of different mimedb specification mechanisms (inline objects, json extension and merging, data urls (ew), etc). Simply saying "we'll use mimes in some way" provides _zero_ details on what the API will actually look like for users (which is either the most important or second most important aspect of the feature) and makes it almost _impossible_ to fairly compare a MIME-based proposal with one of the ones which are not.

bmeck: @weswigham I think the valid possible resolutions of the DB via a string is very different from comparing the API like you are claiming it is. My comment above just applies to what valid resolutions are in the string. The claim of Object vs Array and Delegation vs not are higher level topics that are not in the same level of design.

> Simply saying "we'll use mimes in some way" provides zero details on what the API will actually look like for users (which is either the most important or second most important aspect of the feature) and makes it almost impossible to fairly compare a MIME-based proposal with one of the ones which are not.

I'm not sure what this means. I think you are trying to make some statement that MIME-based proposals significantly differ from other ways of encoding the format of resources, which I don't see anywhere in this PR or other issues.

This PR included details on the APIs that could be used so this comment you made doesn't seem to make sense in context of this PR?

proposals/esm-in-js-ux.md
GeoffreyBooth: I’m thinking of `require.extensions`, where in CommonJS a handler can be registered for a file extension, like `require('coffeescript/register')`. There won’t be something similar in ESM mode?

There shouldn’t be a need for something like `"mimeDBs": ["typescript"]`. There should be a way to use a TypeScript loader without needing this configuration block. The intent of this configuration block is only to override default behavior, including default behavior as specified by a loader.

proposals/esm-in-js-ux.md
GeoffreyBooth: > If you have a JSX format for `.jsx` and wish to disable JSX across all packages

All of these proposals only apply within the current package boundary.

bmeck: @GeoffreyBooth 

> All of these proposals only apply within the current package boundary.

They are applied to a single package boundary but the location of this configuration is important. As @weswigham was getting at, the designs here have far reaching implications.

In particular, if delegation to a shared resource is possible. The application of the configuration can be shared across many packages even if it only applies to specific package boundaries.

michael-ciniawsky: What are the benefits of using MIME in this context? How does MIME solve the main problem about JS parse goals (Script <=> Module)? What happens if a new file type surfaces, would the internal `node` Mime DB always need to be updated to support it? e.g why not something in the direction of

**package.json**
```json
{
   "module": "src/index.coffee",
   "loaders": {
      ".coffee": "coffee-loader",
      ".litcoffee": [ "coffee-loader", { "literate": true } ]
    }
}
``` 

**coffee-loader.js**
```js
export const register = (url) => path.extname(url) === '.coffee')

export function loader () {}
```

instead ?

bmeck: @michael-ciniawsky MIME is just being used as a way to encode the format. So I'm just going to talk about formats, MIME is just the default / standard way of encoding those across systems and is well tested and supported in various places. In particular, `data:` URLs use MIMEs as do HTTP `content-type:` headers to define the format of a resource.

> What are the benefits of using MIME in this context?

The benefits of using formats is that you disassociate your resource from location / consumption based data by keeping the format separate from other data. It allows a resource to define its format without needing to supply contextual data.

> How does MIME solve the main problem about JS parse goals (Script <=> Module)

I think you meant CommonJS and Module? A format can be used to disambiguate these as it can act similar to an `enum` and have all the valid goals that can be loaded and even have the potential for supporting goals that are not supported (like coffeescript).

> What happens if a new file type surfaces, would the internal node Mime DB always need to be updated to support it?

I presume so, that is part of why loading files without extensions is not part of `--experimental-modules`. In particular if we allow loading unknown formats as a specific format such as CommonJS that is problematic because it would be a breaking change to add new file types. In addition this format should be supported for things without file extensions like STDIN.

> e.g why not something in the direction of

> ...

By pairing contextual information like file extension here, we are causing some problems in supporting non-file based approaches. How does this handle things like STDIN / `data:` URLs / etc.

In addition we have an ordering problem, in your example you have paired format declaration and handling in the same chain of delegation. These could have separate orderings and are not tightly coupled in nature.

> `export const register = (url) => path.extname(url) === '.coffee'`

This has a problem for tools that do not ship with a JS VM. Even if loaders require a JS VM to handle resources, declaration of resource formats are able to be defined in a way that they can be statically detected. Routing to an appropriate pipeline for a given format without requiring a JS VM would be ideal.

> `[ "coffee-loader", { "literate": true } ]`

This just looks like an encoding similar to a MIME, however it doesn't work with things like `data:` URLs or HTTP servers.

proposals/esm-in-js-ux.md
ljharb: Not at all - it's talking about how to identify a parse goal to users, and my argument is that "the file extension" is the thing they'll understand the easiest.

proposals/esm-in-js-ux.md
GeoffreyBooth: > They are applied to a single package boundary but the location of this configuration is important. As @weswigham was getting at, the designs here have far reaching implications.

I’m not following. Can you please explain? What implications?

> In particular, if delegation to a shared resource is possible, the application of the configuration can be shared across many packages even if it only applies to specific package boundaries.

What is “delegation to a shared resource”? You mean being able to load the configuration from somewhere, like a package?

What is the need for such complicated configurations that users have a need to abstract them? What’s a use case for having a dozen extension-MIME mappings?

bmeck: > I’m not following. Can you please explain? What implications?

A great example of this is deciding if the configuration allows composition/delegation. Even if configuration is applied within a package boundary, the ability to delegate that configuration allows the final configuration to be coordinated across packages.

> What is “delegation to a shared resource”? You mean being able to load the configuration from somewhere, like a package?

Yup, or URL, or w/e. Just being able to have a package load in the configuration from outside its boundary in a way that other packages could do the same.

> What is the need for such complicated configurations that users have a need to abstract them?

Usability wise it allows managing of many packages and configuring them all at once. "Need" is almost always a bad word since we can get creative with workarounds since we have a turing complete system after all.

> What’s a use case for having a dozen extension-MIME mappings?

Probably none! However, delegation isn't the same as having dozens of extensions. Delegation just allows your configuration to stay in sync across many packages. We do this all the time at my work for things like `eslint` so that we can stay in sync.

proposals/esm-in-js-ux.md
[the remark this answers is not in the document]
ljharb: I don't agree that most people need to know them, ever. Their framework will pick the right mime for them. Users think about the info that makes it into the file: the contents, the name, and the extension.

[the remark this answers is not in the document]
bmeck: That seems to make it better to delegate to the framework rather than listing each file extension in a user's `package.json`.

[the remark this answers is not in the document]
ljharb: MIME only comes into play for a user when they're serving content to a web browser - if they're writing a node-only app, MIME types literally never matter to them, and they never have to know them. I think it would be a mistake to force non-browser-involved node users from having to learn a browser concept just to make their file parse how they want with the file extension they want.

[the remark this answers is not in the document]
bmeck: If we support `data:` URLs MIMEs are the method of encoding format there as well, same for if we support `Blob` APIs.

proposals/esm-in-js-ux.md
GeoffreyBooth: It sounds like you want something like this:

```json
"mimes": "https://my-company-server/somewhere/mimes.json"
```

Or instead of a URL, a path to a file inside that package or in a dependency of that package. The thing is, that’s not really how `package.json` files work. Things like the `dependencies` field, or `scripts` or `engines`, _could_ have shared configuration across multiple packages; but we don’t provide any special functionality for abstracting that. `package.json` files are pretty dumb data configuration files, that only describe the package they’re in. That’s just how they are. If you want to manage them in a controlled way across several packages, well, JSON files are easy to work with programmatically and so you would have to write a script to update all your packages’ `package.json` files accordingly. I feel like this request of “I want `package.json` files to pull part of their configuration from an external source” is a significant increase in scope from the feature that’s being proposed here.

proposals/esm-in-js-ux.md
GeoffreyBooth: Let’s back up for a second. In the current `--experimental-modules`, is there a way for the CoffeeScript package to implement the ESM equivalent of `require('coffeescript/register')`? That is, I could have an `index.mjs` with `import 'app.coffee'` and then `app.coffee` has `import` statements for lots of other `.coffee` files.

michael-ciniawsky:
> **across systems **

:) Since `node` doesn't support `import "https://..."` (HTTP) atm and a particular `node` application normally runs on one/the same system using the file system instead of HTTP transfer, mapping the extension to/sniffing/parsing the MIME is just an extra layer in most cases. Hence I don't (yet) see the benefits for `node` in having an across system compatible format for it's module system in general (but that's separate discussion). 

### `pkg.mime`

By in this context I'm especially referring to (npm) packages which are neither directly consumed via STDIN nor as a `data:` URL (those use cases could use MIME parsing, maybe even passing a Data URL to STDIN `node --eval 'data:<type>[;<encoding>],import x from 'pkg'` potentially solves some `STDIN` related issues 🤔, but that's a separate issue/feature/use case imho), so why would using MIME be a preferable approach for packages compared to e.g `index.js/index.mjs`, `pkg.main/pkg.module` or `pkg.mode` (particularly in terms of setting the Parse Goal (CJS/ESM)? 
e.g 

**package.json**
```js
"mimes": {
   "text/javascript": [ ".mjs" , ".js" , ".coffee", ... ],  // Means ESM (?)
   "application/node": [ ".js", ".coffee" ]                 // Means CJS (?) 
},
```

What happens here in case someone specifies `.js, .coffee`  for 2+ MIME Type mappings?

**package.json**
```js
"mimes": {
   ".mjs": "application/node" // (CJS ?)
   ".wasm": "text/javascript"  
}
```

What happens if someone specifies 'incorrect/confusing' mappings ?

####  `Loaders`

> In addition we have an ordering problem, in your example you have paired format declaration and handling in the same chain of delegation. These could have separate orderings and are not tightly coupled in nature.

Please clearify, I'm sorry but especially "chain of delegation" sounds really vague for someone missing the context on potential implementations for loaders

> This has a problem for tools that do not ship with a JS VM. Even if loaders require a JS VM to handle resources, declaration of resource formats are able to be defined in a way that they can be statically detected. Routing to an appropriate pipeline for a given format without requiring a JS VM would be ideal.

Same to vague and abstract for the average human being 🙃, I don't understand why/how a loader for `node` should work without a JS VM and what other 'appropiated pipelines' you have in mind here (in particular) other then writing a loader for `node`?

> This just looks like an encoding similar to a MIME, however it doesn't work with things like data: URLs or HTTP servers.

Sure, but `data:` URL and HTTP handling are not needed here?

proposals/esm-in-js-ux.md
bmeck: Yes, `node --loader coffeescript/loader index.mjs` works, however you cannot mutate the loader at runtime.

In particular lets make a more concrete case of why mutating the loader at runtime isn't great with ESM:

```mjs
import 'coffeescript/loader';
import 'app.coffee';
```

In ESM, before `coffeescript/loader` evaluates, the entire module for `app.coffee` has already been loaded/instantiated, meaning that its source text has been fetched and turned into a Module Record. Under the default loader `.coffee` is unrecognized so the graph will fail to instantiate. That means that `coffeescript/loader` is never evaluated. All of this is due to the static nature of ESM and isn't really specific to how Node implemented things.

proposals/esm-in-js-ux.md
bmeck: We don't have a plan to support HTTP urls; but that is about what is valid to resolve to, not really a claim against delegation.

> The thing is, that’s not really how package.json files work. Things like the dependencies field, or scripts or engines, could have shared configuration across multiple packages; but we don’t provide any special functionality for abstracting that. package.json files are pretty dumb data configuration files, that only describe the package they’re in. That’s just how they are. If you want to manage them in a controlled way across several packages, well, JSON files are easy to work with programmatically and so you would have to write a script to update all your packages’ package.json files accordingly. I feel like this request of “I want package.json files to pull part of their configuration from an external source” is a significant increase in scope from the feature that’s being proposed here.

We already point to things outside of our package boundary whenever we do bare specifier imports. I don't really understand this? "dependencies" is used to configure a bare specifier to get a resource outside of the package boundary (sometimes even using `git+ssh:` URLs etc.), and can even be a symlink.

In addition, with some people wanting to use package name maps, those too would be configuring potentially remote resources within a package boundary.

I do agree it is bigger in scope than a simple object, but I do think it is more feature complete and is not new in terms of delegation within a package boundary.

bmeck: Ah, forgot to mention, that `"loader":` per design in https://github.com/nodejs/node/issues/18233 also reaches outside of the package boundary, so if accessing things beyond the package boundary is problematic we might need to redesign that.

proposals/esm-in-js-ux.md
bmeck: I think `"mimes": ["jsx"]` is easier than `"mimes": {".jsx": "application/vnd.react.jsx"}`

proposals/esm-in-js-ux.md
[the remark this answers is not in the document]
benjamingr: @ljharb 
> MIME only comes into play for a user when they're serving content to a web browser ...

All this is true, but it doesn't mean we can't use mime types in Node.js for ESM to disambiguate things as users are already familiar with them and they're supposed to mean exactly this (metadata about the file format). What Bradley said about data urls is also interesting.

proposals/esm-in-js-ux.md
GeoffreyBooth: What I meant by `dependencies` was the list of what dependencies are being specified. Like say your app uses `babel-runtime`, `babel-core` and `babel-polyfill` (I’m just making these up). There’s no way to do something like `"dependencies": { "babel-dependencies" }` where `babel-dependencies` is a pointer to those three others. There’s currently no way to have two projects that share their list of dependencies, that pull their list of dependencies from another file, something like:

```json
"dependencies": {
  "my-companys-standard-dependencies": "./deps.json",
  "other-dep-only-used-in-this-app": "^1.0.0"
}
```

Nothing like that is possible today, as far as I’m aware. Maybe there’s a case for adding it, but that’s what I mean by saying that loading the list of MIME mappings from outside of `package.json` feels very out of scope. All of the configuration of a package is currently _in_ the `package.json`. Part of that configuration may reference external things, like `babel-core`, but the configuration itself is inside `package.json`.

bmeck: @GeoffreyBooth I don't understand still. If the complaint is that this doesn't have a 1-1 relationship with how dependencies work, that is going to be true on many levels. Having multiple dependencies however, does fill up a shared namespace similarly to how file extensions are a shared namespace and they also delegate out to external resource. If you would prefer it to be something that exposes flat instead of compositional configuration I would still prefer delegation such as :

```json
{
  "mimes": {
    ".coffee" : "coffescript/mimes",
    ".litcoffee" : "coffescript/mimes",
    ...
  }
}
```

As that still allows sharing and doesn't have the bundled behavior you are worried about, but still allows for sharing. I'm not sure on the API experience I show above being ideal, but the idea of flat and delegating is something we can pursue if composition is not desirable.

proposals/esm-in-js-ux.md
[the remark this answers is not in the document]
ljharb: @benjamingr I agree it doesn't mean we *can't*, i'm suggesting it means we *shouldn't*. data URLs is an interesting point indeed, but I'm not convinced it's enough to force the mental overhead of knowing about MIMEs on users.

devsnek:
the thing here seems to be people can't agree on how node will key module types.

i would argue that we shouldn't be focusing on *where* the map of types goes (a resolve hook could look in a package.json, we don't not need to implement that behaviour) but rather what system we want to use to key the types.

personally i am in favour of mimes.

bmeck:
> so why would using MIME be a preferable approach for packages compared to e.g index.js/index.mjs, pkg.main/pkg.module or pkg.mode (particularly in terms of setting the Parse Goal (CJS/ESM)?

education and consistency.

these have been discussed in many issues and I'm not really wanting to review every API design you post in these threads but can if given enough time.

> What happens here in case someone specifies .js, .coffee for 2+ MIME Type mappings?

We should not use the configuration encoding you have here. In particular it needs to be 1 file extension to 1 format. The usage of arrays in your JSON simply leads to conflicts. Just an API design problem where you should have used a 1-1 dictionary. Since multiple file extensions can map to the same MIME, it needs to be keyed off the file extension.

> What happens if someone specifies 'incorrect/confusing' mappings ?

I'm not sure what this means.

* Invalid MIME string as in it doesn't parse (e.g. `js`)?

I presume it would fail once if it gets routed to format disambiguation in the default loader. Note that userland loaders can supercede their parents, and may still allow those files to work either by not delegating to the default loader or by ignoring any errors.

* Incorrect by typo etc. (e.g. `text/js`)

The default loader would return `text/js` for the format since it parses as a valid MIME. Once it is done going through loaders, Node wouldn't know how to make a Module Record for that format since it is not well known and it would fail there.

For any given format mechanism I would expect similar cases for unparsable values and values that are not known how to turn into Module Records.

> Please clearify, I'm sorry but especially "chain of delegation" sounds really vague for someone missing the context on potential implementations for loaders

Loaders have a well defined order in which they can call each other in various designs. Consider a more concrete case where we have 3 userland loaders:

```sh
node --loader jsx --loader apm --loader code-coverage test.js
```

`code-coverage` mutates the resolution first, then it could delegate up to `jsx`, then `jsx` could delegate to `apm`, which could then delegate to `node` (the default loader).

By tightly coupling formats to resolution semantics we get into an odd situation: which of these loaders formats is the proper one for `node` to return from `resolve()`. In this situation we can somewhat safely say `.js` come from the `jsx` loader and might even be a JSX format. However, what happens if our APM supports JSX:

```sh
node --loader apm --loader jsx --loader code-coverage test.jsx
```

How does `apm` avoid specifing that `.js` files are JSX but still support them?

This is increasingly complicated if people mix usage of `NODE_OPTIONS` and CLI flags since `NODE_OPTIONS` always occurs ahead of CLI flags in our loading process.

Essentially, I'm saying we shouldn't try to tie these configuration mechanisms to the same composition as multiple loaders.

> Same to vague and abstract for the average human being 🙃, I don't understand why/how a loader for node should work without a JS VM and what other 'appropiated pipelines' you have in mind here (in particular) other then writing a node loader?

A runtime loader doesn't need to work without a JS VM, but other preprocessing tools are in our ecosystem as well. It would be ideal if we could support them as long as we don't burden our own implementations. I don't think Node should become a transpiler or anything, but we shouldn't make things more difficult for them if avoidable.

Imagine a costly transform like `webpack`, `prepack`, or `uglify`. Being able to know the format of a given file in the same way as other tools / node would allow unification rather than repeated and different ways of specifying things like `sourceType`.

> Sure, but data: URL and HTTP handling are not needed here?

Consistency matters for both education and allowing systems to avoid creating conflict with others. If we have multiple ways of defining the format of a given resource, that means that we need to keep those multiple format disambiguation methods in sync and keep parity in features between them. There is no clear reason to encourage having multiple different ways to describe the format of a resource, but there are reasons to encourage having a single way to describe the format.

proposals/esm-in-js-ux.md
GeoffreyBooth: It’s not about dependencies. It’s that `package.json` is a single resource that defines the configuration of a package. It lists all the dependencies of a package; it doesn’t link to some other file that lists some or all of the dependencies of a package. `package.json` lists the supported Node engines; it doesn’t link to some other file that has the list of engines. And on and on. `npm` and other tools can read `package.json` and know everything they want to know about the configuration of a package; they don’t need to read `package.json` and then read some _other_ file(s) to find the rest of the configuration. `package.json` is self-contained.

bmeck: Doesnt that same desire to keep it without referencing data outside of the package boundary mean packages should not be able to configure their loaders if the loaders are outside of the package? I admit it is a new concept, but dont see any strong counter argument given that loaders already are doing things crossing package boundaries. Maybe you could explain why formats need to stay within a package if loaders are not able to follow the same in-package.json/same package boundary configuration.

GeoffreyBooth: I don’t see how those concerns are connected. The current design of `package.json`, lockfile aside, seems to be that it’s a single place where all configuration settings of a package can be read from. NPM or other tools don’t need to follow links to other files for additional configuration info, again other than the lockfile. I don’t see how this is related to the concept of a `package.json` file’s configuration _affecting_ code that’s beyond that package’s boundary.

In the case of `mimes` in particular, `mimes` absolutely can’t affect files outside of the package boundary. If I map `.js` to be ESM within my package, I can’t have that affect any apps that depend on my package, or else I might break the app of someone who’s expecting `.js` to be CommonJS. Likewise it shouldn’t affect my own package’s dependents, as I might depend on a package that likewise expects `.js` to be CommonJS. This mapping needs to be very contained within my package. I’m not creating a loader here.

michael-ciniawsky:
hmmm, but for WASM it would definitely not work then, since there is only `application/wasm` (to my knowledge) but WASM Modules aren't spec'd yet. It's either a good argument to make changes to the MIME spec to add something in the direction of `goal=module` and/to give it some traction or otherwise, not to rely on MIME for parse goals at all given the current status quo

ljharb:
Or we could rely directly on extension, since .wasm is pretty clear-cut.

bmeck:
@michael-ciniawsky 

> Some kind of education for the education sake. I'm arguing that the what to educate is fairly important here and simply disagree with the currently proposed contents of this proposal, since there are imho 'better' (simpler, more explicit) ways available

I don't see them being explained? Which 'better' ways are you talking about, and how do they fix the problems I've listed in the thread above. In particular, how are they going to keep parity when all other designs are being done using MIMEs and APIs are even using MIMEs (e.g. `data:` and `Blob` usage).

> *Confusing as it is theoretically possible to override any extname with a custom MIME Type via package.json which is not needed as the main issue is still how to deteremine the Parse Goal for .js files and mainly only for them (atm) (e.g what would be other cases?). MIME also doesn't directly provide information about a parse goal for a certain Content-type, so assuming/using application/node (CJS) and text/javascript (ESM), especially the latter is not correct (text/javascript;goal=module would be, but that doesn't exists (yet))

I think you are getting a bit confused, the Script goal of JS is never checked for. You can serve `<script type=text/javascript src=index.html>` and it simply ignores the MIME. MIME is not used in [classic script loading per the WHATWG spec](https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-classic-script) except to use the [charset parameter that exists in all MIMEs](https://tools.ietf.org/html/rfc2046#section-4.1.2). Currently the JS MIME only ever refers to the Module goal of JS in terms of standards for web browsers.

> Plus pkg.mime has the same limitations then pkg.mode adding a 'package boundary' which makes it impossible to provide a 'dual-mode' package without having to rely on different file extensions 🤷‍♂️

This is a limitation that seems fine to me as most people seem to be able to work only in one format or be able to use multiple file extensions? Use a [per package `"loader":`](https://github.com/nodejs/node/issues/18233) for more complex behaviors and routing of files?

> The ordering problem is a common problem for any plugin/loader system. At some point one simply has to know when a particular plugin/loader needs to be executed [in relation to other plugins/loaders]. This can only be mitigated and due to that fact it is always the fundamental question if introducing such a system solves more issues then the intrinsic complexity adds

I don't understand if this is for or against keeping the configuration points decoupled. It seems like it is neither for or against keeping things coupled so I'm just going to state that it seems simple to keep them decoupled and means you don't have to execute JS to determine such things.

> In the context of node loaders, this would still require a node loader which e.g spawns a child_process to execute and handle a non-JS tool etc? Integration needs to happen somewhere, otherwise it is an enterily separate process

Yes, but this is not necessarily being done when the application runs. Loaders can be run ahead of time such as if someone wants to do things like a transpiler loader or resolve all URLs to absolute URLs ahead of running an application. The claim here is that there isn't really any advantage to 

> hmmm, but for WASM it would definitely not work then, since there is only application/wasm (to my knowledge) but WASM Modules aren't spec'd yet. It's either a good argument to make changes to the MIME spec to add something in the direction of goal=module and/to give it some traction or otherwise, not to rely on MIME for parse goals at all given the current status quo

WASM modules are first class already [in terms of language design](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Module). There doesn't need to be any special MIME for them so I'm a bit confused at this comment. All that is being discussed is how to properly make [facades](https://docs.google.com/presentation/d/14Jn4ry0rlONMd2Yc1ZdFl-aU42XKVEgOc9x2Hy9C9ZI/edit#slide=id.g3c73ede8b6_0_28) for them to integrate with ESM since the current [Abstract Module Record](https://tc39.github.io/ecma262/#sec-abstract-module-records) and [Source Text Module Record](https://tc39.github.io/ecma262/#sourctextmodule-record) constraints do not satisfy the needs of WASM integration.

proposals/esm-in-js-ux.md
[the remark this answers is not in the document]
bmeck: I'm trying to point out that, whatever we do, we need to be able to encode it as a MIME for any **API** that uses MIMEs, and thus taking the position that it is more overhead to not learn MIMEs than it is to learn MIMEs. We don't tell people we shouldn't use RegExp when trying to do string based pattern matching because RegExps are more complex than globs, the idea of MIMEs being a mental overhead seems to fall in the same category.

proposals/esm-in-js-ux.md
bmeck: > I don’t see how those concerns are connected. The current design of package.json, lockfile aside, seems to be that it’s a single place where all configuration settings of a package can be read from. NPM or other tools don’t need to follow links to other files for additional configuration info, again other than the lockfile. I don’t see how this is related to the concept of a package.json file’s configuration affecting code that’s beyond that package’s boundary.

This isn't true, `node_modules` configures all of the bare specifier paths of a given scope, without being in `package.json` itself. In addition things like [Per Package Loaders](https://github.com/nodejs/node/issues/18233) would configure the package's loading system in `package.json` by delegating to a shared resource.

Is there a specific reason all of this data should be in one place when people already need to defer to `node_modules` and will need to defer to Loader implementations (global and per package) to understand the package already?

What are we design space trying to preserve??? Education seems possible, but we already have to make multiple lookups to understand things outside of `package.json` so what makes MIMEs special and needing to avoid the same sort of multiple location lookup.

bmeck:
Just to clarify to everyone. It was always my intention in [my gist mentioned for mappings](https://gist.github.com/bmeck/7ee7eb2147e2dafe3167c856d9b4151a) that such databases not only be built into Node but also to be configurable by users and tools.  I'd be fine if there was both the `[...databasePaths]` that looked up a resource and allowed composition/delegation and an alternative that allowed an in-band `{"extension": "format"}` form. Is there anyone thinking that we should limit ourselves to just one of these 2? The `[]` form from my gist that looks up paths to databases/includes a few reserved ones is my preferred one and others have a preference for the `{}` form, but both should be able to be statically analyzed by tools or by human lookup.

If we could amend the proposal to perhaps have both forms of this as I don't think they conflict with each other?

We should probably document removing existing mappings from being delegated to like `{".node": null}` or w/e for the `{}` form as well to help keep some parity if that sounds ok as well.

GeoffreyBooth:
I’m trying to clarify @bmeck’s proposal and how it allows user-specified mappings. How about something like this:

Assume a new `mimes` section in `package.json` that can be an object, e.g.:
```json
"mimes": {
  "js": "text/javascript",
  "cjs": "application/node"
}
```

That `mimes` section could alternatively take an array of strings, which would be relative paths to JSON files:

```json
"mimes": [
  "./jsx.json",
  "./typescript.json"
]
```

Each JSON file would be an object with “extension: mime” mappings like the first example. They would be combined using `Object.assign`, e.g. something like:

```js
const packageJsonMimes = JSON.parse(fs.readFileSync('./package.json')).mimes;
let mimes = {};
if (Array.isArray(packageJsonMimes) {
  packageJsonMimes.forEach((filePath) => {
    const mimeMappings = JSON.parse(fs.readFileSync(filePath));
    Object.assign(mimes, mimeMappings);
  }
} else {
  mimes = packageJsonMimes;
}
return mimes;
```

And that’s it. The idea is that not just Node but also build tools need to be able to parse this `mimes` section, and so it can’t be terribly complicated. We can’t have `"mimes": ["typescript/mimes.json"]`, I don’t think, because then a build tool would need to be able to implement Node’s package name resolution algorithm in order to find that `mimes.json` file. That wouldn’t be difficult for Node to do, of course, but it would be a challenge for a build tool—especially one not written in JavaScript—to achieve.

@bmeck is this a fair characterization of what your proposal would look like in practice? Everyone else, is the utility of defining these mappings in external files worth the additional complexity, especially if the external file can’t be inside a dependency?

ljharb:
@GeoffreyBooth it wouldn't be challenging at all; every build tool already uses something like https://npmjs.com/resolve to do just that. More to the point, if it allowed *any file path whatsoever*, it'd have to be a require path imo, otherwise it would make no sense.

bmeck:
@GeoffreyBooth 

> @bmeck is this a fair characterization of what your proposal would look like in practice? Everyone else, is the utility of defining these mappings in external files worth the additional complexity?

It also included the idea of removing mappings and a null database to ensure no default mappings are applied, so some things would change slightly, to be more composed like:

```mjs
const packageJsonMimes = JSON.parse(fs.readFileSync('./package.json')).mimes;
let mimes = Object.create(null);
if (Array.isArray(packageJsonMimes) {
  for (const filePath of packageJsonMimes) {
     if (filePath === null) break;
      const mimeMappings = JSON.parse(fs.readFileSync(filePath));
     Object.assign(mimes, mimeMappings);
  }
} else {
  mimes = packageJsonMimes;
}
return mimes;
```

> We can’t have "mimes": ["typescript/mimes.json"], I don’t think, because then a build tool would need to be able to implement Node’s package name resolution algorithm in order to find that mimes.json file.

I'm not sure how valuable delegation is if it cannot be in a shared package.  `"./node_modules/typescript/mimes.json"` etc. are fragile and lead to other systems like meteor rewriting that kind of import/require when they encounter it. Since `typescript` can be above the current package on disk even if it is a dependency.

A lot of places already have the node resolution algorithm in place, Java had it for Closure Compiler/Ocaml for Flow/JS via the lib @ljharb mentioned/etc. have libraries and if package name maps were introduced those tools would need to implement that resolution algorithm as well.

If it only is for relative paths I don't think it is valuable enough to warrant being shipped since it would lead to people using those fragile paths instead of encouraging tools to use libraries that implement the node resolution algorithm that already exist for various languages.

GeoffreyBooth:
> Since `typescript` can be above the current package on disk even if it is a dependency.

This presents a problem though. If `typescript` is some globally installed package, then the behavior of the current package varies based on the environment it’s in. Like if I have TypeScript 2.0.0 installed globally and it has some MIME mappings defined, but then my colleague has TypeScript 3.0.0 installed globally and 3.0.0 ships with different MIME mappings, then the package with `"mimes": ["typescript/mimes.json"]` in it behaves differently on my machine from my colleague’s.

bmeck:
> If typescript is some globally installed package, then the behavior of the current package varies based on the environment it’s in.

Globally installed packages are not part of resolution.

GeoffreyBooth:
> Globally installed packages are not part of resolution.

Then what do you mean by “`typescript` can be above the current package”?

ljharb:
@GeoffreyBooth something like a node_modules like this:
```
foo
  |----- typescript
  |----- bar
         |--- typescript
         |--- baz
```
where both foo and baz do `require('typescript')`

GeoffreyBooth:
Okay, if you’re telling me that there won’t be resolution issues, like for something like the NPM website trying to read the package to list stats about it on its webpage, then sure.

@bmeck what’s the use case for `null`? Why would we want to start from a blank slate rather than Node’s defaults?

bmeck:
@GeoffreyBooth main use case is disabling formats that node would have by default. `.json`, `.node`, and CJS are not importable on the web natively, so setting them to be errors/not found would ensure they are not used in a package.

GeoffreyBooth:
Okay, so how about this:

___

Assume a new `mimes` section in `package.json` that can be an object, e.g.:
```json
"mimes": {
  "js": "text/javascript",
  "cjs": "application/node"
}
```

That `mimes` section could alternatively take an array of strings, which would be relative or resolved paths to JSON files:

```json
"mimes": [
  "./my-mimes.json",
  "typescript/mimes.json"
]
```

Each JSON file would be an object with “extension: mime” mappings like the first example. They would be combined using `Object.assign`. If the first element in the array is `null`, Node’s default MIME mappings are discarded first before the array elements are merged, e.g. something like:

```js
const nodeDefaultMimeMappings = require(‘module’).mimeMappings;

const packageJsonMimes = require(‘./package.json’).mimes;
if (Array.isArray(packageJsonMimes)) {
  let mimeMappings = (packageJsonMimes[0] === null) ? Object.create(null) : nodeDefaultMimeMappings;
  packageJsonMimes.forEach((filePath, index) => {
    if (filePath === null && index !== 0) {
      throw new Error(‘Only the first element of a mimes array may be null’);
    }
    Object.assign(mimeMappings, require(filePath));
  }
  return mimeMappings;
} else {
  return Object.assign(nodeDefaultMimeMappings, packageJsonMimes);
}
```

___

It doesn’t make sense to allow `null` in any array element position but the first, does it? If it were second, say, I would assume it would wipe out everything that came before it, in which case why would the user put something before it that just gets erased.

bmeck:
@GeoffreyBooth the ordering matters depending on which takes precedence and if you have a mental model of the array delegating to node's defaults at the front or at the end.

`NODE_DEFAULTS.concat(mimes)`
vs
`mimes.concat(NODE_DEFAULTS)`

I tend to have a mental model of `null` being a terminator and the code being similar to `mimes.concat` instead of `NODE_DEFAULT.concat`. Perhaps just because implicit data I tend to think of as being at the end of the list rather than potentially at the beginning. In that thinking I think it would make sense for `null` to be at the end of the list.

I don't have strong preference on restricting things such that they can't be at one end of the list. In particular I'm unsure about how people will want to concatenate/merge these lists. I'd be fine restricting it to one end for now, but would just like a mental model to go through, and for it to match any `--loader` ordering.

https://github.com/nodejs/node/pull/18914 introduced multiple loaders and treats things as right to left and delegating at the far right to the "default" loader. This is to allow `NODE_OPTIONS` to be preempted by CLI arguments, so I think your thinking that processing this might actually be better going right to left as well for the mental model. That would mean that `null` should be at the beginning of the list.

whew... @GeoffreyBooth yes the above is about right.

GeoffreyBooth:
@bmeck Regarding the presets accumulating left to right or right to left, I assume the docs for this will say something like “the objects will be combined as if by an `Object.assign` call,” and `Object.assign` works left to right. So I think that makes more sense of the two options.

I’m going to update the PR with this option, and it sounds like we’ve found consensus!

ljharb:
@GeoffreyBooth note that I still think that it should be a map of extensions, not mimes, (and i think some others on the thread agreed) so i think that detail still needs to be hashed out.

GeoffreyBooth:
@ljharb I reread the comments and you seem to be the only one in favor of using file extensions as a way to signal Node how to interpret a file. Several others wrote that it’s confusing to map extensions to extensions.

Only six people commented on this thread, so we can discuss with the larger group at the meeting.

GeoffreyBooth:
@devsnek, @zenparsing, @sendilkumarn, thanks for offering to contribute! I created a fork [here](https://github.com/GeoffreyBooth/node) and invited you to contribute to it. If anyone else would like to contribute, please let me know and I’ll be happy to add you.

Again, this is _a parallel effort_ to the new fork that @MylesBorins is spearheading. If the group would consent to removing `--experimental-modules` from core while we work on a new fork, I would be happy to shift my focus to the new fork/new implementation. But as long as `--experimental-modules` is being released to the public for end users to use, I feel an urgency to improve it, especially to [make it compatible with CoffeeScript](https://github.com/nodejs/modules/issues/151) and other tools that have similar compatibility issues. I expect that the new fork will probably end up using a lot of code from `--experimental-modules`, including whatever code we write for this PR, so this work will likely not be wasted effort.

I created a new branch [`esm-in-.js-files`](https://github.com/GeoffreyBooth/node/tree/esm-in-.js-files) to hold the work to implement the above proposal. The branch is a copy of @guybedford’s [`package-mode`](https://github.com/guybedford/node/tree/package-mode) branch that was the basis for https://github.com/nodejs/node/pull/18392, because that branch had a lot of code in it that I presume will be needed by the PR we hope to open. I’m thinking that the basic path forward will be to edit the work done in that branch to change the implementation from `"mode": "esm"` to the `"mimes"` section as defined by the top of this thread. @devsnek and anyone else more familiar with the Node codebase, does this seem about right to you?

I understand that @MylesBorins has concerns about how this interacts with loaders. I’m hoping that that’s not something we need to figure out for the purposes of implementing this PR; that we can go with whatever support that https://github.com/nodejs/node/pull/18392 had for loaders, and leave it at that. My impression from talking to others in the group is that the loaders implementation in `--experimental-modules` is incomplete (again, @devsnek or others, please chime in if that’s incorrect) and I’m thinking that fully fleshed out loader support should be something added as its own PR, that would likely end up building upon some of the work in this PR. As in, I’d rather not this PR get blocked by implementing loaders, if we can achieve this PR without tackling all that work.

Does this seem like a workable plan? Thanks!

devsnek:
@GeoffreyBooth 

> thanks for offering to contribute

Just to be clear, I'm willing to help out with questions you have about the innards, (e.g. mentor, like I said above). I'm not really in a good place to be taking on additional projects right now.

> My impression from talking to others in the group is that the loaders implementation in --experimental-modules is incomplete.

It is incomplete, but it is designed very intentionally to support certain directions and plans we had/have so I would advise you to be careful. This code is very tightly coupled and you can introduce subtle (mostly timing related) bugs if you aren't careful.

sendilkumarn:
@GeoffreyBooth  👍  Is there any other medium for communication?

GeoffreyBooth:
@sendilkumarn You can use my name at gmail to reach me via email or Hangouts.

GeoffreyBooth:
@guybedford Do you mind sharing any insights of how you’d recommend going about adapting your PR into the proposal at the top of this thread?

guybedford:
@GeoffreyBooth I'm not sure, but if it would help to hear, the original conception here was that `"mode": "esm"` referred to the default ESM "MIME database with .js files as ESM. And custom modes like `"mode": "coffee"` or something like that would allow opting into custom treatment per-package, where such a mode database might be defined by a `--loader` or similar. This wasn't fleshed out at all though, so the above is merely illustrative.

MylesBorins:
I'm going to remove this from the agenda until we can reach consensus on a minimal kernel / approach for the fork. Please feel free to re add or let me know if this should be on today's agenda

proposals/esm-in-js-ux.md
```diff
@@ -0,0 +1,116 @@
+These are proposals for the user experience of an addition to `package.json` that configures Node to treat `.js` files as ESM within a package boundary. This assumes the `--experimental-modules` implementation as a baseline. This is spun off from [#150](https://github.com/nodejs/modules/issues/150), and relates to [nodejs/node/pull/18392](https://github.com/nodejs/node/pull/18392).
+
+### [nodejs/node/pull/18392](https://github.com/nodejs/node/pull/18392)
+
+This pull request adds support for a `"mode": "esm"` flag to be added to `package.json`, that tells Node to treat all `.js` files within the package boundary of that `package.json` file as ESM. The package boundary of a `package.json` file is deemed to be all modules in that folder and subfolders until the next nested `package.json` file.
```
jkrems: Does this imply that every `import 'x/y/z.js'` would necessarily trigger a search for `package.json` files along the path hierarchy?

ljharb: I'd expect so, that's how the current require implementation, and the ecosystem, works.

MylesBorins:
Is this something we want to continue to try and land or should we close it?

MylesBorins:
Is this something we want to continue to try and land or should we close it?

GeoffreyBooth:
I think let’s please keep it open for now? We’re referring to the `"mimes"` proposal in the Stage 2 proposals, and it could very well end up getting implemented in the new modules implementation too. This may have been written in the context of `--experimental-modules`, but configurability of Node’s treatment of file extensions is potentially useful in any implementation.

bmeck:
@Alhadis MIME includes parameters since its [first RFC](https://tools.ietf.org/html/rfc1341). However, note that compliant usage of MIME requires that unrecognized parameter names be ignored. For `charset` it is likely we will parse it, but if we comply with browsers for ESM it *must* be UTF-8 per https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-single-module-script which ignores the charset and decodes as UTF-8 always (and the same for all other script types).

Alhadis:
That's all well and good for ESM, but when authors start plugging their own loaders and formats into the module system, it's harder to guarantee authors will always be using UTF-8... 😞

devsnek:
Most of the JS ecosystem assumes UTF-8. in node you should be able to use buffers and text decoders to handle the difference.

bmeck:
@Alhadis as @devsnek mentions, any custom loaders can convert to supported formats such as JIS conversion to UTF-8 before handing it off to the runtime.

Alhadis:
Ah that's different. Sorry, I thought the MIME mapping was attempting to address custom loaders as well. My bad.

bmeck:
@Alhadis custom loaders should be able to query these, but for a variety of scenarios we cannot tie them together. E.g. if a package supports `coffeescript` and `iced-coffeescript` generates `coffeescript` you encounter a very good example scenario:

1. Package `is-iced` could list the MIME for `iced-coffeescript`.
2. A custom loader could generate `coffeescript` from those `iced-coffeescript` files.
3. A custom loader generates `javascript` from those `coffeescript` outputs.
4. The runtime knows how to load `javascript` and does the appropriate steps.

We have nested compilation in that scenario and it shows why you need loaders to generate the appropriate formats that the runtime supports by default. If you hand `node` a `coffeescript` file, it doesn't know what to do with it, it needs that loader to convert it to something that it understands. The same is true for character encodings.

GeoffreyBooth:
Wow, an Iced CoffeeScript reference. @bmeck you warm my heart.

@Alhadis This proposal is based on the currently-shipping `--experimental-modules` implementation, and we’re building a new implementation with a different design over in the [ecmascript-modules](https://github.com/nodejs/ecmascript-modules) repo. A road map of its development is [here](./doc/plan-for-new-modules-implementation.md). As currently designed the new implementation doesn’t have a need for user configuration of how Node handles file extensions, at least for treating `.js` files as ES modules, but this proposal is still here in case we decide to add such functionality in either implementation at some point.

Fishrock123:
My 2c: `mode` (with common named modes) is incredibly more ergonomic to average user understanding than `mimes`, and I would prefer `mode` like that any day. (Even if it solves all problems slightly less well.)

GeoffreyBooth:
One other thought: we could even have `mode` _and_ `mimes`, to provide the user with lots and lots of control, if we can work out how the two would interact. This almost starts to turn `package.json` into a loader via configuration, but that’s not necessarily a bad thing.

MylesBorins:
soooo an idea. What if "loaders" and "mode" were the same thing.... we have a number of built in loaders that are document e.g `esm`, `cjs`, `json`, `native`.  Then we could have a table such as

```js
loaders: {
".js": "nodejs:cjs",
".esm": "esm",
".mjs": "nodejs:esm",
".wasm": "@loader-central/wasm",
".ts": "./loaders/typescript.js"
}
```
As you can see here we have a combination of built-in loaders `nodejs:loader`, loaders from modules `esm`, loaders from scopes `@loader-central/wasm`, and loaders from local files system `./loaders/typescript.js`

Thoughts?

devsnek:
I like this direction, especially being able to specify loaders.

ljharb:
Except for the builtin module part being a url instead of a scope (but assuming that we’d stick with whatever node core decided), that looks great to me. (I’d expect all builtin loaders to be under some kind of `loader/` path, as well)

guybedford:
I like the idea of the progressive enhancement, but would want to keep the string value as the Phase 2 approach, where basically the string value effectively determines exactly the full map without having to carefully specify each extension, which most users shouldn't need to do I don't think.

Also, are we sure we want to make it easy to publish uncompiled source formats to npm? The module translator buffet still feels like a slippery slope to me, that we might want to leave to loaders in general.

devsnek:
@guybedford you don't need to specify every combo, just the ones you want to override. additionally `"mode": "esm"` doesn't tell me *anything* about what it is doing. Is it making all files load as source text? is it making `.esm` files work? is it running my code with jdd's `esm` package? it just doesn't have good ergonomics.

bmeck:
@guybedford nothing is preventing people from requiring a global loader to compile their code properly to run. I've stated in the past being able to statically know your loaders is preferable as well which global loaders do not have since they are just coming from the CLI/ENV. I like that this approach does let people see what loaders are expected and allows tooling to do things with that knowledge statically.

I do have a list of features I would expect regarding multiple points of composition like the array example above.

1. The ability to declare a specific format as "unhandled"/maps to an unknown format (browsers treat this as the "" type)

```json
{
  "loaders": {".node": null}
}
```

Disabling default handling of ".node", return "" (or some well known signifier) as the expected format.

2. The ability to intercept *ALL* files. Thinking of `"loaders":` as an Array would satisfy this where each item can map to arbitrary file extensions, or a "catch all".

```json
{
  "loaders": [
    {".ts": "always-typescript", ".json": "always-json"},
    "always-cjs"
  ]
}
```

3. The ability to use a short string to summarize some common usage. `"loaders": "nodejs:default"` for example might apply to multiple file extensions so it needs to be a catch all which seems to have some conflict with `"nodejs:cjs"`.

4. It should be easy to configure things without writing a bunch of JS yourself. This leads to lots of JS events and piping data around through loaders... I think this needs to be thought out more to minimize spinning up loaders just to manage format disambiguation based upon file extension.

devsnek:
@GeoffreyBooth btw there is more down here based on some additional constraints which isn't reflected in the "consensus" section at the top.

GeoffreyBooth:
@devsnek Sure, I think at this point we would need to start a new thread or a new repo like one of the recent proposal repos. The consensus here was based on the currently-shipped `--experimental-modules`, and so a feature based on our new implementation might very well look different. Also per https://github.com/nodejs/modules/pull/160#issuecomment-454953764, we might want to design this after or in tandem with loaders.

But still, there was a lot of thought put into the discussion here, so this should be at least a starting point I think for whatever we continue working on.

GeoffreyBooth:
Closing as this was based on the old `--experimental-modules` and there are new issues and PRs created since this discussion ended.

Issue #119, opened 2018-05-23
Initiative: Terminology / Historical Decisions documents


SMotaal:
**Motivation**: In the last few days, while playing catch up to the things I missed in the last 3 weeks, I caught a glimpse of how challenging it would be for our future selves or others going through this repository's issues and documents when trying to pull together various threads.

**Initiative**: I'd like to propose and take part in an effort to put together *terminology / historical decision documents* which will allow us to keep track of easy-to-misread terms and how or why certain technical terms will evolve.

**Volunteers:** @smotaal @devsnek @bmeck @guybedford 

**Tasks:**
- [X] Pick Format: GD
- [X] Determine Initial Structure: taxonomy-like
- [X] [Initial Draft](https://docs.google.com/document/d/1khaKKB1id0A52moNA9gbXE9bvV8Sisyr-IhC4i9dQdQ/edit?usp=sharing) (@smotaal)
- [X] Create [terminology branch](https://github.com/nodejs/modules/tree/terminology) (@smotaal)
- [X] Create Terminology.md #158

smotaal:
@MylesBorins should I add the modules-agenda label?

benjamingr:
Good initiative! 

This should cover static/dynamic resolution, how named exports work, live bindings, differences between ESM and CJS, the various proposals in the past (summarized), where to find what in the repo (links to the use cases for example).

Would you be willing to take charge of it?

smotaal:
@benjamingr I understand the pain more than anyone at this point not to mention that I have a love-hate relationship with terminology.

It would be a lot more effective to have one or two other volunteers to increase the odds of it making sense to humans.

bmeck:
I vote that we use a doc unless there is good reason to have version control. Not the biggest fan of wikis on github.

GeoffreyBooth:
I was about to nominate a wiki, because it allows Markdown. I find the use cases/features docs hard to read because the code is all mangled in Google Docs.

bmeck:
@GeoffreyBooth idk, comments are nice and allow issues to be done without needing to spin up a full github thread about specific terms, also it allows a nicer sharing mechanism that github wikis. if formatting is a concern we might be in big trouble since this is a list of terminology and collection of links and not a book with full on examples I would think. If we want to do full examples of each term with code highlighting that might a bigger goal than what I thought this was and would probably need its own project.

GeoffreyBooth:
[...]

PR #164, opened 2018-08-08
Package & Module sub-terms for Agnostic Consumers

robpalme:
Add sub-terms of _Agnostic Consumers_:

- ***Agnostic Package Consumers***
- ***Agnostic Module Consumers***

The immediate reason for adding these terms is to help separate the debates around:

- interop across package boundaries - which maybe less controversial
- interop within a package - which appears to be controversial

I'd welcome feedback on whether different terms would be clearer, e.g. perhaps ***inter-package interop*** vs ***intra-package interop***

### DIFF
--- a/doc/Terminology.md
+++ b/doc/Terminology.md
@@ -17,6 +17,28 @@ Normative Terms
   This term alone does not specify in which direction(s) the agnosticism applies.
 </dd>
 
+<dt id=agnostic-package-consumers>
+  Agnostic Package Consumers
+</dt>
+<dd> 
+  A restricted form of [Agnostic Consumers](#agnostic-consumers) that applies only to imported modules referenced via a `package.json` file, as opposed to directly.
+  
+  This implies the imported resource is the main entrypoint into another package.
+</dd>
+
+<dt id=agnostic-module-consumers>
+  Agnostic Module Consumers
+</dt>
+<dd> 
+  A restricted form of [Agnostic Consumers](#agnostic-consumers) that applies only to imported modules referenced directly, as opposed to via a `package.json` file.
+  
+  This implies the imported resource either:
+  
+  - does not reside within any package, or
+  - resides within the current package, or
+  - is a deep import into another package, i.e. it bypasses the default entrypoint resolution process
+</dd>
+
 <dt id=require-interop>
   Require Interoperability <code>require(<esm>)</code>
 </dt>


demurgos:
This issue is about the different possible import specifiers. Following the various discussions and proposals, I think that clarifying it is important but it may more complicated.

Here are the different cases where the proposals can diverge:
- relative specifier (starts with "./" or "../") VS bare specifier
- simple bare specifiers VS deep bare specifiers: "pkg" VS "pkg/deep"
- resolved module outside the package boundary or not: Even with relative path, you can use things like "./foo", "../../../../outside/foo", "./node_modules/foo" (is it inter or intra package?)
- specifier pointing to a directory containing a package.json or not (more relevant than simple bare VS deep bare? eg. "@babel/types" has a package.json: you're not grabbing individual files inside, unlike "@angular/animations/browser")


I had a comment somewhere else that I find "agnostic import" more fitting than "agnostic consumer" because the "lack of knowledge of the dependency implementation" is defined for each individual import (static ES import / dynamic ES import / CJS require), not at the module level (and to me "consumer" strongly implies the "consumer module").
Framing the terminology around imports (instead of consumers) would allow to more naturally use qualifiers describing the specifier: "agnostic relative import", "agnostic deep bare import", "agnostic extra-package import".

At the moment, Node has no concept of package boundaries. It seems that the most relevant distinction is if the specifier points to a file, a directory with "package.json" or a directory without "package.json". The relative/bare distinction just decides if we walk the parent directories looking for "node_modules" or not. Since the most common use case for specifiers pointing to a file are "intra-package" imports and for specifiers pointing to directories with "package.json" are "inter-package" imports, I'm fine with using it as a shortcut, but it must be clearly understood as so.
Defining the terminology with "package" may also be imprecise in the case where a file is a standalone script without any `package.json` around: are its relative imports "intra-package" despite there being no package at all?

ljharb:
Please let’s not use the word “import” to mean anything except “the keyword used inside ESM”

ljharb:
All the more reason to avoid overloading the word. Loading is fine but connotes loaders specifically; i think “consuming” might work.

demurgos:
https://github.com/nodejs/modules/issues/119#issuecomment-406192845

> I consider the term "import" to be ambiguous because it depends on the context. I see two possible interpretations:
> - **ES import**: A static `import` statement or dynamic `import()` as defined by the ES spec.
> - **CJS or ES import**: Either an ES import or a `require` (represents the abstract concept of importing something regardless of the concrete import mechanism)

I agree with you that it's also a point causing confusion, but I don't feel "loading" or "consuming" are the best terms to mean "any kind of import mechanism". It probably deserves a separate issue, but in the context of my previous comment I used it to mean "any kind of import mechanism". I prefer people to clearly say "ES import" or "CJS require" when the difference matters, and just "import" when the difference does not matter. I think this fits better with the terminology I propose:
- "agnostic package ES import": A static or dynamic `import` for a package, without knowledge of its implementation
- "agnostic file import": Any kind of of import (ES import or CJS require) for a file or simple directory, without knowledge of its implementation.

robpalme:
My summary so far on the feedback from @demurgos @ljharb @bmeck is that:

- the ***package*** & ***module*** qualifiers and their definitions have no objections; in fact @demurgos comes to the [same conclusion](https://github.com/nodejs/modules/pull/164#issuecomment-411337980)
- we're not sure what noun to pick:  ***consumer*** vs ***importer*** vs ***loader***

So far I agree with @ljharb that using _import_ in this context is misleading because it falsely implies we're talking exclusively about the `import` or `import()` forms. 

Whilst I don't think we need to reserve the term `load` exclusively for things that involve _loaders_, in this case where we are using _loader_ as a noun, it's highly ambiguous, e.g. is **agnostic package loader** referring to _a module that loads packages agnostically_, or is it referring to _a loader that deals with packages agnostically_?

Hence I come back to the original **agnostic consumer** as the best way to talk about something that loads things agnostically.

SMotaal:
@robpalme was this included in #158 before it was finally merged? Can we close this if so?

GeoffreyBooth:
This has conflicts, but I’m not sure we need this PR at this point. Can this be closed?

robpalme:
@GeoffreyBooth At this point I think the conversation has moved onto concrete proposals so this refinement of terminology is no longer needed.

PR #180, opened 2018-09-12
doc: first pass at minimal-kernel

MylesBorins:
Decided to submit this as a PR to comment and land as documentation.

Thoughts?


doc/minimal-kernel.md
```diff
@@ -0,0 +1,41 @@
...
+  - no native modules
+  - createRequireFunction used to get these
+  - will come back with format database
+
+# How will we implement the minimal kernel
+
+* removing importing of formats other than ESM
+  - no common.js
+  - no JSON
+  - no nativemodules
```
michael-ciniawsky:
```diff
- no nativemodules
+ no native modules
```

doc/minimal-kernel.md
```diff
@@ -0,0 +1,41 @@
...
+* removing of dynamic path searching
+  - no extension adding
+  - no directory resolution
```
michael-ciniawsky:
```
- no directory resolution
- no support for index.js
```

Isn't that the same ?

```
- no directory resolution (dir/ => dir/index.js)
```

MylesBorins:
updated based on feedback.

Biggest changes

* expand minimal kernel
  - added dynamic import
  - added import.meta.url
* documented PRs that implement the kernel
  - https://github.com/nodejs/ecmascript-modules/pull/2
  - https://github.com/nodejs/ecmascript-modules/pull/3
  - https://github.com/nodejs/node/pull/19360

doc/minimal-kernel.md
```diff
@@ -0,0 +1,44 @@
...
+* we cannot have dynamic path searching
+  - requiring the full path is an issue for tooling and a long term solution is required
+  - migration strategies also have issue with this
+* common.js backwards compat
+  - createRequireFunction does
+  - import.meta.require does not fail early enough
+  - hold off on importing common.js until more progress is made on dynamic modules spec
+* .mjs will be the only way to import ESM
+  - intention is to move forward with format databases to map extensions and support multiple use cases
+* only supporting importing ESM
```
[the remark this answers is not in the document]
SMotaal: Liking this *explicit evantuality* theme... it matches with the intent of moving towards somewhat expected features (after reaching consensus obviously)

MylesBorins:
Updated significantly based on feedback I received from @GeoffreyBooth 

As such I'm dismissing all current reviews. PTAL

SMotaal:
Amazing!

doc/plan-for-new-modules-implementation.md
```diff
@@ -0,0 +1,50 @@
...
+These features will be part of the first phase of development:
+
+* `module.createRequireFromPath` ([nodejs/node#19360](https://github.com/nodejs/node/pull/19360)) is the only way to import CommonJS into an ES module, for now.
+  - `import.meta.require` fails at runtime as opposed to import time. This is not desireable to all committee members
+  - Hold off on `import` statements for CommonJS until more progress is made on the dynamic modules spec.
+  - landed in https://github.com/nodejs/node/commit/246f6332e5a5f395d1e39a3594ee5d6fe869d622
+
+* `import` statements will only support files with an `.mjs` extension, and will import only ES modules, for now.
+  - In a later phase, the intention is to move forward with format databases to map extensions and support multiple use cases.
+  - No JSON or native modules; `createRequireFunction` can be used to get these.
```
mcollina: We should clarify if we _might_ ship that in the future here.

targos: Isn't it what the preceding sentence says?

MylesBorins:
ping @nodejs/modules

WebReflection:
> `import` statements will only support files with an `.mjs` extension, and will import only ES modules, for now.

If in the later phase the intention is to move forward with format databases to map extensions and support multiple use cases, this implementation has no reasons to accept only `.mjs` as extension, and it fights against the principle of Phase one:

> if Node is doing something that browsers also do, Node should do it in the same way. Alternatively, code that executes in both environments should produce identical results.

If using `import` is the way to disambiguate the developer intent, **let it go** and make `.js` as ESM possible too.

WebReflection:
> So let's "bank" the precious consensus we have, draw a line under it by merging this PR, and then progress the additional capabilities in Phase two. You can still get what you want ;-)

I'm all for that, but I really would like to underline, after all this time, that the consensus is basically the following one:

  * no CJS interoperability out of the box, you need to opt in to load CJS from ESM
  * nothing magic in terms of resolution, it's either relative or absolute path

These two points means that to have CJS there is an explicit opt in, so that everything else written with ESM syntax is meant to load ESM, as it's been on browsers land for years now, and same goes to other JS oriented projects that supports ESM already.

Accordingly with Phase one, there is absolutely no reason whatsoever to force an extension that was born prematurely to bring in interoperability out of the box, because you want CJS, you have to explicitly do that, so that `.js` in ESM's phase one can only mean **JavaScript** if imported via `import`.

I hope by now we all understood that interoperability out of the box was **the issue**, never the solution so that `.mjs` doesn't belong to this Phase one which is, otherwise, an awesome starting point.

As summary, since we agreed we cannot have CJS in ESM out of the box, and we shouldn't indeed, cannot we let the extension we all know and use already work right away and have in Phase one the right core that doesn't introduce breaking/mandatory changes to the rest of the ecosystem (browsers) that worked already 'till today?

'cause indeed everything else can be done after, including the ability to use `.mjs`, if developers, bundlers, whoever, like that.

MylesBorins:
@WebReflection some context you may be missing about the minimal-kernel as I don't believe you were in the meetings. The point of the kernel is to be compatible with all potential future iterations. We could not reach consensus on any defaults. There was consensus that `.mjs` would stick around, even if we support `.js` for esm in the future. As such, supporting only `.mjs` in the first phase was the only way to allow all future updates to be non Semver-Major.

After this lands we can create proposals that change the defaults, but it is IMHO out of scope for the minimal kernel

WebReflection:
> The point of the kernel is to be compatible with all potential future iterations.

That's the **only** reason I've stated with phase one, which has zero ambiguity, it makes no sense to rule out the most natural extension for JavaScript, which is `.js`.

But as you asked me in private, I'm happily stop following this group and will never directly interfere if not explicitly asked to.

Best Regards.

jdalton:
In the PR based on this doc PR, https://github.com/nodejs/ecmascript-modules/pull/6, I noticed that this document is missing a stance on including/excluding `--loader` and `--experimental-vm-modules` support. 

It could be interpreted that the lack of a mentions in the 
> ## Phase 1: The Minimal Kernel
> These features will be part of the first phase of development

section means they aren't included.

I think there may be folks that want to experiment in the loader space and I believe providing one up-front will guide folks down a specific route _(not great)_. I don't see the `--loader` option as mission critical for a minimal barebones implementation.

Similarly, I see `--experimental-vm-modules` as extra and not mission critical for a minimal barebones implementation. That's not to say anything needed for internal minimal functionality couldn't stay, just that the user facing bits can probably be dropped.

porsager:
> > The point of the kernel is to be compatible with all potential future iterations.
> 
> That's the **only** reason I've stated with phase one, which has zero ambiguity, it makes no sense to rule out the most natural extension for JavaScript, which is `.js`.
> 
> But as you asked me in private, I'm happily stop following this group and will never directly interfere if not explicitly asked to.
> 
> Best Regards.

I've been following this (silently) for quite some time, and I think it's a pity that the viewpoints here are silenced since I'm very sure they actually represent a lot of users. If the arguments/suggestions @WebReflection has aren't valid or impossible to obtain, shouldn't it be possible to describe why that is, or link to some place it has been described before? I haven't been able to find such refutation, so if anyone has a place to read up on it for me (and other silent readers) it would be great.

MylesBorins:
@porsager 

from this document

> The general idea is to start with a “minimal kernel” as Phase 1, which consists of features that the @nodejs/modules group have agreed will be necessary for all potential iterations of our ESM implementation

The purpose of the minimal kernel is not to handle every use case but to enable every potential implementation. @WebReflection specified the following earlier

> If using import is the way to disambiguate the developer intent, let it go and make .js as ESM possible too.

This is not accurate. We do not have consensus that import will 100% be unambiguous, it is possible that transparent interop for CJS may still happen in a future iteration. If we allow `.js` to be used for ESM in the kernel we would have to make a breaking change. There are a number of people working at TC39 right now to help design a way to import CJS modules with named exports in a spec compliant way.

What we do have consensus on is that all potential iterations of our future module implementation will include `.mjs` as a unambiguous extensions to reference an ESM module. As such this feature made sense to include in the kernel.

Anyone will be able to propose an enhancement to the kernel and we should be able to move from that design space to a completed implementation in iterations without breaking anything.

michael-ciniawsky:
> This is not accurate. We do not have consensus that import will 100% be unambiguous, it is possible that transparent interop for CJS may still happen in a future iteration.

This can also easily be flipped and interpreted that there is also no consensus that it **will** actually be ambiguous in the end

> If we allow .js to be used for ESM in the kernel we would have to make a breaking change.

Which breaking change? In regard to `--experimental-modules` shipped in `node`?

> There are a number of people working at TC39 right now to help design a way to import CJS modules with named exports in a spec compliant way.

Wasn't the conclusion more or less that this doesn't need to be adressed by TC39 directly and instead `node` and V8 should work out and implemenation instead? Are dynamic modules === backwards compat or is there a 'benefit' for ES Modules aswell? Is adoption by browser vendors realistic/likely? What is the estimated time frame/deadline here?

> What we do have consensus on is that all potential iterations of our future module implementation will include .mjs as a unambiguous extensions to reference an ESM module. As such this feature made sense to include in the kernel.

It may make sense to use it **for now** and finally start iterating on an implementation, but it's still a very controversial addition which recieved **a lot** of community pushback (which is superficially acknowledged, but actionably **ignored** as it seems) and with a debatable 'technical benefit' (for node) compared to the potential downsides (if it has one at all). E.g the initial premise of using `import x from './'` and to distiguish between ESM (`.mjs`) and CJS (`.js`) (Transparent Interoperability (TI)) via file extensions is also kind of outdated, since extension and directory resolving and TI were removed (for now) and following the discussion I can't recall that under these conditions the need for a new file extension was reevaluated here. Therefore of the the iteration phases of the minimal kernel implementation are only **additive** in terms of features, then supporting a new file extension should be treated more carefully as it actually is avoidable, if some kind of TI isn't feasible (which seems to be the current status)

medikoo:
While I understand why (at this starting phase) support of ESM is supposed to be limited to .mjs, it also appears to me that it contradicts raised _Browser equivalence_ goal.

When I write `import foo from ./foo.js` in module I expect it'll work same in Browser and Node.js, and current plan doesn't guarantee it'll ever be the case.

It appears it's either we keep the door open for _.mjs only_ ESM support in Node.js __or__ ensure _Browser equivalence_.

Both can't be achieved at once, and it'll be good to have this transparent

guybedford:
@WebReflection I'm just trying to understand what you are asking for - are you saying you want `.mjs` removed from this minimal implementation entirely, instead having it just supporting `.js` for ES modules? Then for us to work out the ambiguity problem separately, assuming we will solve it somehow?

WebReflection:
@guybedford I am saying that since phase one has no ambiguity, there is no reason to force users to use an extension born to solve ambiguity. However, since I understand some user might have used `.mjs` at this point, I think both `.js` and `.mjs` extensions should be accepted.

**As Summary**, I'd like the following point to be consistent with the scope, and the goal, of this phase one:

* `import` statements will support both files with a `.js` or `.mjs` extension, and will import only ES modules, for now.

And keep everything else untouched, which is otherwise excellent as it is.

All considerations about how to explicitly and eventually map `.js` to CommonJS runtime, can be done later on. At this point, with this phase goal and agreements, there is no ambiguity to solve, hence there is no reason to load ESM only via `.mjs`.

I hope it's clear now, and thanks for consideration.

Best Regards

MylesBorins:
The kernel is never going to be shipped and was not designed thinking about
UX first because of that

Getting hyper focused on specific missing fearures completely misses the
point of this excersize

WebReflection:
> Getting hyper focused on specific missing fearures completely misses the point of this excersize

I thought the point of this exercise was to move forward with what has been agreed and preserve what browsers already shipped. Why is this then focused on using the `.mjs` extension which is not needed at phase one and doesn't respect what browsers shipped already?

Why isn't `.mjs` not considered an extra feature but a must have, while the natural extension for JavaScript, and all browsers, has always been `.js`?

If the purpose of the micro kernel is to work phase after phase on agreements, who agreed on the need for `.mjs` at phase one and disagreed that `.js` shouldn't work?

Why is it so hard to admit there's no reason to explicitly mention `.mjs` at this phase one as it is?

MylesBorins:
While browser equivalence is considered at this stage 100% compat was not the goal for phase one.

As the first paragraph of this PR States we only include features that so not block potential future features. Earmarking .js for esm blocks our ability to use it for CJS in a future phase and as such was removed from phase 1.

WebReflection:
> Earmarking .js for esm blocks our ability to use it for CJS in a future phase and as such was removed from phase 1.

That means `.js` has been reserved without consensus, right?

But this PR states the following:

> In a later phase, the intention is to move forward with format databases to map extensions and support multiple use cases.

The `.js` to CJS should eventually be mapped later on, not virtually reserved without consensus, IMO.

But I guess it's just me seeing this sneaky little detail paving the wrong path from the scratch again.

Anyway, we are going in circles and I've nothing else to add.

Best Regards.

MylesBorins:
I think that is out of scope for Phase 1 but could be proposed as an
enhancement afterwards. Since phase one is not optimized for user
experience and isn't shipping it seems out of scope

rubys:
@MylesBorins increasing consensus is out of scope?

> Earmarking .js for esm blocks our ability to use it for CJS in a future phase and as such was removed from phase 1

I'll note that earmarking __.mjs__ for esm blocks our ability to use it for CJS in a future phase and so why isn't that ALSO removed from phase 1?

Put another way, earmarking at this stage of the game seems unnecessarily controversial.  So why do it?  Note: I'm assuming that `--experimental-modules` itself goes away eventually, so this is emphatically **not** a user experience proposal.

The intent of my question is to explore the possibility of deferring locking in a decision that is clearly controversial at this point.  I think that is entirely in the spirit of phase 1.

bmeck:
I'd like to note, if we have people wanting to remove loading `.js` for reservation and people wanting to remove `.mjs` for reservation... we effectively have nothing.

MylesBorins:
> I'll note that earmarking .mjs for esm blocks our ability to use it for CJS in a future phase and so why isn't that ALSO removed from phase 1?

The challenge here is the .js is ambiguous and can be used for any potential javascript goal. The browser has the ability to deal with this via out of band information (mimes), we do not.

Even if we utilize `.js` in the future for ESM, the extension remains ambiguous forever. `.mjs` was proposed, and has ben used, as an unambiguous way to refer to ESM. This is being standardized at [the ietf](https://tools.ietf.org/html/draft-ietf-dispatch-javascript-mjs-02) and we are attempting to get support for it in well used mime databased such as [python](https://github.com/python/cpython/pull/3908).

The consensus reached in meeting was that we had agreement on `.mjs` as an extension for unambiguous ESM, and that consensus was not going to change. As such it was the perfect candidate for extension reserving.

It may be a bit harsh, but it seems to me that the only people this decision has been controversial to are individuals who have not been actively engaged in the process. I mentioned a timeline above, we've been actively discussing this kernel for close to 6 weeks. At a certain point we need to move forward with something, anything.

mathiasbynens:
I honestly don’t understand the pushback. The PR _literally_ addresses the desire to support `.js` and other extensions:

> In a later phase, the intention is to move forward with format databases to map extensions and support multiple use cases.

Phase 1 is minimal _by design_ — the doc itself says as much. Can we please stop suggesting more work items for the phase 1 milestone, and stop blocking progress on modules in Node.js altogether?

rubys:
> It may be a bit harsh

Yes, I do see that as harsh.  I read it as "I'm new to the party, and therefore I can't play".

I've been quietly listening, looking for an opportunity to slow down and eventually reverse the trend towards Node.js making and locking down paternalistic and opinionated choices on behalf of the node development community, and instead look to [leave decisions such as these to the users](https://github.com/nodejs/modules/issues/185).

I'll also note that my question was not answered.  I'd like to see the community make room for people who wish to participate by quietly listening and asking thoughtful questions.

MylesBorins:
> I read it as "I'm new to the party, and therefore I can't play".

That's absolutely not my intent, I'll step away from the conversation for a bit as I'm not doing a good job of expressing myself. For the record I see a big difference between new people wanting to engage and those who have been members for a while but not active.

> Node.js making and locking down paternalistic and opinionated choices on behalf of the node development community

I don't find this particular assessment of Node entirely accurate. The ideology of small core is very much about making as few decisions as possible, but ensuring they are the correct ones.

To your initial question

> What would break if --experimental-modules accepted a parameter which contains an extension (or possibly a list of extensions). Specifically, what would break if someone specified --experimental-modules=js?

I'm not entirely sure tbqh, and I think this would make an excellent proposal in phase 2.

jkrems:
nit: Do we need to call out support for importing core modules? Or is that implicitly "what the current core implementation does"?

MylesBorins:
@jkrems I've updated the doc to be explicit about the support for built-ins

I've also updated https://github.com/nodejs/ecmascript-modules/pull/6 to remove VM + loaders... thanks @jdalton for the code

GeoffreyBooth:
Regarding the `.mjs` mention in Phase 1, I don’t think there’s disagreement that `.mjs` can be used as an unambiguous way to signify that a file should be parsed as ESM; it seems to me that people’s objections are that it might end up as the _only_ way, as it is now with `--experimental-modules`. Hence the mention of the file extension-parse goal mapping in Phase 1, even if it gets implemented in a later phase. I’m okay with this privileged place for `.mjs` in Phase 1 in the interest of moving forward, since we won’t finish this new modules implementation without providing a way for ESM to be used in `.js` files.

Regarding “breaking change,” I interpreted @MylesBorins to mean “between Phase 1 and later phases.” Maybe I’m wrong on that, and if so please correct me. Assuming I’m correct in my interpretation, I think a better phrase could have been chosen. Since we’re not publishing Phase 1, there are no users for whom Phase 2 would break anything. I understand the desire to plant seeds in Phase 1 such that future phases are purely additive, but I’m sure it won’t work out quite that perfectly. That’s okay. Even the consensus that’s formed in early phases is subject to change in later phases. Even that, though, is _progress,_ settling on something we can agree on, even if it’s only for now.

MylesBorins:
We had consensus in a meeting with quorum landing!

Thanks everyone

mroderick:
> It may be a bit harsh, but it seems to me that the only people this decision has been controversial to are individuals who have not been actively engaged in the process. I mentioned a timeline above, we've been actively discussing this kernel for close to 6 weeks. At a certain point we need to move forward with something, anything.

If you have specific individuals in mind, perhaps you should invite them to participate in "the process".

If you don't like questions and pushback on ideas by community members, which certainly slows everything down, perhaps these discussions should only be open to comment on by members of the `nodejs` organisation?

robpalme:
@mroderick In yesterday's public call of the Modules group, @MylesBorin invited engaged observers to step up and volunteer to become active members.

That meeting was broadcast live on YouTube and the recording is available to view now. I d encourage you to watch it, not just because it might address your concern, but also because it was the most productive meeting we've had so far 🙂

MylesBorins:
> If you have specific individuals in mind, perhaps you should invite them to participate in "the process".

As Rob mentioned anyone is able to raise an issue to join the team. I was specifically speaking to people already on the team who are not actively participating.

PR #200, opened 2018-10-13
add @soldair to observers

soldair:
hey folks. o/
I'm working on defining a common base for user land loaders. My goal is to make it easy to make loaders that speed up node startup time, and work well together. 

I'm excited to contribute and it was great meeting some of you at the collab summit.


MylesBorins:
+1

MylesBorins:
consensus to accept @soldair into the team.

Welcome welcome!

SMotaal:
Finally, and welcome aboard @soldair... This is a good time to overload with all the threads in this effort. In case you are there already, I think that along with the first few new issues and this [comment](https://github.com/nodejs/modules/pull/196#issuecomment-432817177) and PR #196 generally that is where things are happening as of late.

PR #215, opened 2018-11-05
Add format distinction to phase 2

guybedford:
I'd like us to add the format distinction question to the Phase 2 process.

It seems important to finally tackle this one and get some movement on solutions here.

This includes any top-level --module flag, as well as internal imports to CommonJS for switching from ESM loading to CJS loading.


guybedford:
PR approved in today's meeting.

PR #218, opened 2018-11-07
governance: be explicit about meetings new members can join

MylesBorins:
Be explicit that if a new members has requested to join and the issue has no objections that can attend the meeting where their membership is approved


bmacnaughton:
Thanks for working on clarifying this. I didn't want to presume the right to show up.

I'm completely outside the process, so I might be missing some assumptions. This change sounds like it is focused on new members where the membership approval has largely been determined prior to the meeting making it official. I would not have been sure that this applied to observers had I read it before the meeting.

If the intention is that members means members and observers then it should probably be explicit.

PR #242, opened 2018-12-22
doc: update stage 2 regarding dynamic modules

MylesBorins:
As agreed upon in Dec 19th meeting

### DIFF
--- a/doc/plan-for-new-modules-implementation.md
+++ b/doc/plan-for-new-modules-implementation.md
@@ -70,8 +70,7 @@ These changes are implemented in https://github.com/nodejs/ecmascript-modules/pu
 
 * Implement specification changes related to dynamic module records
   - Proposal: ["Dynamic Modules Proposal"](https://github.com/nodejs/dynamic-modules/)
-  - We will need to reach consensus on appropriate behavior for [`export * from 'dynamic-module'`](https://github.com/nodejs/dynamic-modules/pull/11).
-    The current behavior of throwing may be reverted at a later time.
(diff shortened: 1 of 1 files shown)


doc/plan-for-new-modules-implementation.md
```diff
@@ -68,6 +68,10 @@ These changes are implemented in https://github.com/nodejs/ecmascript-modules/pu
 * Define semantics for determining when to load sources as CommonJS or ES module for both the top-level main (`node x.js`) and dependency loading.
   - Proposal: [“File Specifier Resolution” proposal](https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal) covers `import` statements of ESM files; and CommonJS files, package entry point and package deep imports.
 
+* Implement specification changes related to dynamic module records
+  - Proposal: ["Dynamic Modules Proposal"](https://github.com/nodejs/dynamic-modules/)
+  - We will need to reach consensus on appropriate behavior for [`export * from dynamic module`](https://github.com/nodejs/dynamic-modules/pull/11)
```
guybedford: Can we add another point here please to indicate that dynamic modules should not block Phase 2 from shipping?

Something like:

- If consensus is not reached, dynamic modules will be carried over into a further phase.

jdalton: On the added line for reaching consensus:
Would it make sense to add a blurb that throwing for `export * from dynamic module` isn't necessarily intended as a permanent addition and that, at a later time, the behavior could be reverted? Or is that better off in the dynamic-modules proposal?

MylesBorins:
If anyone want to add to this please feel free to push to the branch 😇

PR #280, opened 2019-02-27
Revise road map per current status

GeoffreyBooth:
Readable version: https://github.com/GeoffreyBooth/modules/blob/update-roadmap/doc/plan-for-new-modules-implementation.md. The diff is a mess so you might want to just read that side-by-side with the [current version](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md).

@MylesBorins and I worked together to update our new implementation’s plan per [the Google doc where the team has noted our current status](https://docs.google.com/document/d/1DSWrdV1fzXvlOdTZ5MngDX7v6CU4ZUheJ7ysOZ2uK0w/edit?ts=5c63dc98). In particular:

- The “How will we get from where we are to Phase 1” section has been moved to a new Phase 0, to reflect that it’s already done and was a precursor to our Phase 1.
- The end of Phase 2 corresponds with replacing the existing `--experimental-modules` with our new implementation, while the end of Phase 3 corresponds with dropping the `--experimental-modules` flag. There is no longer an empty Phase 4.
- Phase 2 includes the features listed under the “Requirements to replace upstream implementation” section in the [Google doc](https://docs.google.com/document/d/1DSWrdV1fzXvlOdTZ5MngDX7v6CU4ZUheJ7ysOZ2uK0w/edit?ts=5c63dc98). Previous Phase 2 features that weren’t in that section of the Google Doc have been moved into Phase 3.
- Phase 3’s features are split into two groups: UX improvements, and things that need consensus.

This is just a rewrite of the document to _reflect_ the plan as it stands in the Google doc. This PR isn’t intended to _change_ the plan. If the group decides to move features between phases, that can certainly happen, but we didn’t want to include plan changes along with reformatting this document.

### DIFF
--- a/doc/plan-for-new-modules-implementation.md
+++ b/doc/plan-for-new-modules-implementation.md
@@ -76,26 +76,28 @@ Phase 2 fleshes out the implementation with enough functionality that it should 
   - Proposal: [“File Specifier Resolution” proposal](https://github.com/GeoffreyBooth/node-import-file-specifier-resolution-proposal) covers `import` statements of ESM files; and CommonJS files, package entry point and package deep imports.
   - Landed in https://github.com/nodejs/ecmascript-modules/pull/28.
 
-* Define semantics for enabling ESM treatment of source code loaded via `--eval`, STDIN, and extensionless files.
+* Define semantics for enabling ESM treatment of source code loaded via `--eval`, STDIN, and extensionless files (both with and without shebang lines).
   - Proposal: [“Entry Points Proposal”](https://github.com/geoffreybooth/node-esm-entry-points-proposal) covers non-file forms of input as well as adding `--type` flag for controlling file-based input.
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
I revised this per our agreements in the 2019-02-27 meeting. Per that meeting this will land soon unless there are any objections.

MylesBorins:
I would move reaching consensus on file extensions to phase 2 based on the current discussion

GeoffreyBooth:
You mentioned a target date when we want to have a PR ready to submit to core for upstreaming? Do you know what that date is and should we add that to this? (Can be a future PR.)

MylesBorins:
@GeoffreyBooth maybe March 13th? I don't think we are blocked by technical implementation at this point. We can have the out of band meeting about extension resolution, get remaining work landed, and prepare a branch that we plan to upstream. We can then get consensus on landing it in the next meeting.

PR #281, opened 2019-02-27
doc: update members list with ncu-sync

MylesBorins:
Updated based on github IAM roles


giltayar:
Myles moved me down to an observer role at my request.

I have full confidence in this group, and had a wonderful experience. Unfortunately, I don't have the time to dedicate to this group in a way that is meaningful.

I will stay on as an observer (as I have been practically doing for the last few months), and hopefully will be able to communicate some of what is going on to developer outside this group.

PR #287, opened 2019-03-04
Move loaders to phase 3

guybedford:
This PR moves the new loaders work to Phase 3.

This is assuming https://github.com/nodejs/ecmascript-modules/pull/47 to include _a_ solution in the mean time.

I've worked with @bmeck to understand where the current implementation is, and did consider driving this work over the next week to get something together, but we should probably try to be sensible about the design process here.

### DIFF
--- a/doc/plan-for-new-modules-implementation.md
+++ b/doc/plan-for-new-modules-implementation.md
@@ -96,7 +96,6 @@ Phase 3 improves user experience and extends the MVP. Phase 3 is malleable based
 
 * A loaders solution that supports all items in the [features list in our README](https://github.com/nodejs/modules/#features).
   - Should loaders be per package, per application or either?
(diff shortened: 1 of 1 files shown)


doc/plan-for-new-modules-implementation.md
```diff
@@ -98,6 +94,10 @@ Phase 3 improves user experience and extends the MVP. Phase 3 is malleable based
 
 ### UX Improvements
 
+* A loaders solution that supports all items in the [features list in our README](https://github.com/nodejs/modules/#features).
+  - Should loaders be per package, per application or either?
+  - Will land in Phase 2 only if an implementation without major problems (e.g. memory leaks) can be completed in time. If the problems can be isolated behind a flag specific to loaders, we could upstream a buggy implementation and unflag it after its bugs are fixed.
```
GeoffreyBooth: Do we still need this bullet point now that this is in Phase 3?

guybedford: Thanks, I've removed that.

PR #288, opened 2019-03-04
Update Phase 2 file extension/directory index searching plan

GeoffreyBooth:
Update per today’s meeting, includes #287.


GeoffreyBooth:
Both PRs received preliminary consensus to land per the meeting, will leave here for two days before landing in case anyone has any objections to the specific wording.

Issue #255, opened 2019-01-22
JSON Modules


MylesBorins:
I've opened this up on the whatwg/HTML repo.

https://github.com/whatwg/html/issues/4315

Trying to figure out steps to supporting importing JSON

MylesBorins:
rough implementation in https://github.com/nodejs/ecmascript-modules/pull/43

PR #291, opened 2019-03-13
Announcement blog post

GeoffreyBooth:
Here’s a draft of what our announcement blog post could be.

https://github.com/nodejs/modules/blob/announcement/doc/announcement.md


doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+The previous `--experimental-modules` allowed `import` statements of JSON and native modules. This has been removed; you may use `module.createRequireFromPath()` for these.
+
+There is ongoing work in the standards space to cover JSON and WASM and other future potential module types. Node.js will add support for these in compliant ways when the standard is finalized.
+
+## ES module code in packages
+
+_This is a work in progress and subject to change._ You can create a package with ES module sources by using the `package.json` `"main"` field to point to an ES module package entry point. Node will know to load it as an ES module if the file ends in `.mjs` or if the `package.json` also contains `"type": "module"`.
+
+Currently it is not possible to create a package that is usable both via `require` (as CommonJS) and via `import` (as an ES module). There are efforts underway to address this, and may involve changes to the above (especially concerning `"main"`). Please do not publish any ES module packages intended for use by Node.js until this is resolved.
```
jkrems: Given this, do we want to talk about the details here, e.g. even mention using `main` above?

devsnek: i think so. we want people to use these features, just not to depend on them in published packages.

jkrems: Right, but this feature is specifically about published packages. So what would "using them" mean in this case if not using it in a published package?

devsnek: you can use a package.json without publishing to npm :shrug:

jkrems: True, but I'm not sure how common that is. My impression so far is that people are actively surprised when they are told that you *can* use `package.json` to influence file resolution without involving installed (or linked) packages.

GeoffreyBooth: > Given this, do we want to talk about the details here

I wanted to shy away mentioning the proposals under consideration, to avoid seeming to prefer any of them (especially if we end up going with something not yet proposed) and because that felt like too much detail. I thought it was worth mentioning `"main"` specifically to tell users that the UX they will interact with might change, that we’re not just changing implementation details under the hood.

I’m very open to changing this paragraph if you’d like to suggest something better, I was just trying to prevent another `"module"` situation where the community started relying on something that wasn’t standardized.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
+# Announcing a New `--experimental-modules`
+
+Back in 2017, Node.js 8.9.0 shipped experimental support for [ECMAScript modules](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/), known for their `import` and `export` statements. This support was behind the flag `--experimental-modules`.
+
+A lot has happened since then. All major browsers [now support](https://caniuse.com/#feat=es6-module) ECMAScript modules (ES modules) via `<script type="module">`. Various projects have sprung up to make NPM packages with ES module sources available for use in browsers via `<script type="module">`. Support for [import maps](https://github.com/WICG/import-maps), which bring to browsers Node-style package names in `import` statements, [is coming to Chrome](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/qYeQFqgFOyA/rXJapjMaEAAJ).
```
ljharb: always lowercase npm everywhere; "ES module" → "ES Module"; "node" is lowercase unless it's "Node.js", i think?
```suggestion
A lot has happened since then. All major browsers [now support](https://caniuse.com/#feat=es6-module) ECMAScript modules (ES modules) via `<script type="module">`. Various projects have sprung up to make npm packages with ES Module sources available for use in browsers via `<script type="module">`. Support for [import maps](https://github.com/WICG/import-maps), which bring to browsers node-style package names in `import` statements, [is coming to Chrome](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/qYeQFqgFOyA/rXJapjMaEAAJ).
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+All this progress toward adoption of ES modules has increased the urgency for Node.js to ship its support for ES modules. Especially now that there are other runtimes and environments where ES modules are in use, it’s more important than ever that Node.js join the JavaScript standard.
```
ljharb:
```suggestion
All this progress toward adoption of ES Modules has increased the urgency for Node.js to ship its support for ES Modules. Especially now that there are other runtimes and environments where ES Modules are in use, it’s more important than ever that Node.js join the JavaScript standard.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+Yet Node’s initial ES modules support has remained in an experimental state, largely due to strong pushback from the community to several decisions that were made in its design. In response, the [Modules Team](https://github.com/nodejs/modules) was formed to take a fresh look at implementing support for ES modules in Node.js. This has led to a [new implementation](https://github.com/nodejs/ecmascript-modules) for supporting ES modules, that we are pleased to announce will ship as part of Node.js 12. It will replace the old `--experimental-modules` implementation, behind the same flag. We hope that this new implementation addresses many of the community’s concerns, and can ship as part of Node.js proper, without a flag, before Node.js 12 reaches [LTS status](https://nodejs.org/en/about/releases/) in October 2019.
```
ljharb:
```suggestion
Yet node’s initial ES Modules support has remained in an experimental state, largely due to strong pushback from the community to several decisions that were made in its design. In response, the [Modules Team](https://github.com/nodejs/modules) was formed to take a fresh look at implementing support for ES Modules in Node.js. This has led to a [new implementation](https://github.com/nodejs/ecmascript-modules) for supporting ES Modules, that we are pleased to announce will ship as part of Node.js 12. It will replace the old `--experimental-modules` implementation, behind the same flag. We hope that this new implementation addresses many of the community’s concerns, and can ship as part of Node.js proper, without a flag, before Node.js 12 reaches [LTS status](https://nodejs.org/en/about/releases/) in October 2019.
```

ljharb: > largely due to strong pushback from the community

I don't think this is entirely accurate. Certainly there was pushback, but I don't believe that's the reason it remained experimental prior to the WG getting spun up. I think this part should be severely tweaked or removed.

(Especially considering ".mjs by default" is one of the things that got the largest pushback, and that has *not* changed)

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+## What’s in `--experimental-modules`
+
+Like the previous version, this new `--experimental-modules` adds support to Node.js for the following:
+
+- [ES2015 `import` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) can reference JavaScript files that are themselves written using ES module syntax. Files can be referenced as relative URLs (`'./file.mjs'`), absolute `file://` URLs (`'file:///opt/app/file.mjs'`), package names (`'es-module-package'`) or paths within packages (`'es-module-package/lib/file.mjs'`).
```
ljharb:
```suggestion
- [ES2015 `import` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) can reference JavaScript files that are themselves written using ES Module syntax. Files can be referenced as relative URLs (`'./file.mjs'`), absolute `file://` URLs (`'file:///opt/app/file.mjs'`), package names (`'es-module-package'`) or paths within packages (`'es-module-package/lib/file.mjs'`).
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `import` statements that reference ES module files can specify both the default export (`import _ from 'es-module-package'`) or named exports (`import { shuffle } from 'es-module-package'`). All Node.js built-in packages like `fs` and `path` support both default and named exports.
```
ljharb:
```suggestion
- `import` statements that reference ES Module files can specify both the default export (`import _ from 'es-module-package'`) or named exports (`import { shuffle } from 'es-module-package'`). All Node.js built-in packages like `fs` and `path` support both default and named exports.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `import` statements that reference CommonJS files (all current JavaScript code written for Node.js, [using `require` and `module.exports`](https://nodejs.org/api/modules.html)) can use the CommonJS default export (`import _ from 'commonjs-package'`) only. _This is a work in progress and may change in the future._
+
+- [`export` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) in ES module files can specify both default and named exports for `import` statements to reference.
```
ljharb:
```suggestion
- [`export` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) in ES Module files can specify both default and named exports for `import` statements to reference.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Dynamic `import()` expressions can be used to import ES modules into either CommonJS or ES module files. Note that `import()` returns a promise.
+
+- `import.meta.url` provides the `file://` URL of the current ES module file.
+
+- Node.js can be run with an ES module file as a program’s initial entry point.
```
ljharb:
```suggestion
- Node.js can be run with an ES Module file as a program’s initial entry point.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `import` statements that reference ES module files can specify both the default export (`import _ from 'es-module-package'`) or named exports (`import { shuffle } from 'es-module-package'`). All Node.js built-in packages like `fs` and `path` support both default and named exports.
+
+- `import` statements that reference CommonJS files (all current JavaScript code written for Node.js, [using `require` and `module.exports`](https://nodejs.org/api/modules.html)) can use the CommonJS default export (`import _ from 'commonjs-package'`) only. _This is a work in progress and may change in the future._
+
+- [`export` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) in ES module files can specify both default and named exports for `import` statements to reference.
+
+- Dynamic `import()` expressions can be used to import ES modules into either CommonJS or ES module files. Note that `import()` returns a promise.
+
+- `import.meta.url` provides the `file://` URL of the current ES module file.
```
ljharb:
```suggestion
- `import.meta.url` provides the `file://` URL of the current ES Module file.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `import` statements that reference ES module files can specify both the default export (`import _ from 'es-module-package'`) or named exports (`import { shuffle } from 'es-module-package'`). All Node.js built-in packages like `fs` and `path` support both default and named exports.
+
+- `import` statements that reference CommonJS files (all current JavaScript code written for Node.js, [using `require` and `module.exports`](https://nodejs.org/api/modules.html)) can use the CommonJS default export (`import _ from 'commonjs-package'`) only. _This is a work in progress and may change in the future._
+
+- [`export` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) in ES module files can specify both default and named exports for `import` statements to reference.
+
+- Dynamic `import()` expressions can be used to import ES modules into either CommonJS or ES module files. Note that `import()` returns a promise.
```
ljharb:
```suggestion
- Dynamic `import()` expressions can be used to import ES Modules into either CommonJS or ES Module files. Note that `import()` returns a Promise.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `import.meta.url` provides the `file://` URL of the current ES module file.
+
+- Node.js can be run with an ES module file as a program’s initial entry point.
+
+- Loaders can be written to modify Node’s runtime behavior with respect to ES modules. _This is still very much a work in progress._
```
ljharb:
```suggestion
- Loaders can be written to modify Node’s runtime behavior with respect to ES Modules. _This is still very much a work in progress._
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+So if that’s what’s carried over from the previous implementation, what’s new?
+
+## `import` and `export` syntax in `.js` files
+
+Yes, we heard [you](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md). And [you](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054). And [you](https://twitter.com/maybekatz/status/1062473765865512961). You can now use `import` and `export` syntax in `.js` files.
```
ljharb: This seems a bit disingenuous, since that wasn't prevented before - it just was not, and still is not, the default.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+The `.mjs` extension is still supported, and still always explicitly means that its file uses the Module parse goal (and therefore is an ES module). We also created a new `.cjs` file extension, to explicitly signify that a file should be treated as CommonJS. (CommonJS is the other module system that Node.js supports, with `require` and `module.exports`.) These two extensions allow individual files to define how they should be loaded, as ES modules or as CommonJS.
```
ljharb:
```suggestion
The `.mjs` extension is still supported, and still always explicitly means that its file uses the Module parse goal (and therefore is an ES Module). We also created a new `.cjs` file extension, to explicitly signify that a file should be treated as CommonJS. (CommonJS is the other module system that Node.js supports, with `require` and `module.exports`.) These two extensions allow individual files to define how they should be loaded, as ES modules or as CommonJS.
```

ljharb: I still object to the `.cjs` extension - I feel it was snuck in and not properly discussed in a meeting - and I do not want it mentioned *in any way* publicly, since I intend to have it removed.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+As it turns out, a lot of people didn’t like needing to use a new file extension for JavaScript. Browsers don’t require it, tools that build and process JavaScript have issues with it, and languages that transpile into JavaScript such as TypeScript and CoffeeScript don’t necessarily know when to save their output as `.js` or as `.mjs`. So while `.mjs` and `.cjs` have use cases that only they can solve (ES module and CommonJS files side-by-side in the same folder, loose JavaScript files outside of any package) the modules group searched for solutions other than `.mjs` to let Node.js know when to interpret a file as an ES module.
```
ljharb: Re OP: this entire paragraph.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+This flag provides a way to support ES module syntax for input via `--eval`, `--print` or `STDIN`.
+
+## Mandatory file extensions
+
+In the new `--experimental-modules`, file extensions are mandatory in `import` statements: `import './file.js'`, not `import './file'`. However, the CommonJS-style automatic extension resolution behavior (`'./file'`) can be enabled via a new flag, `--es-module-specifier-resolution=node`. (Its inverse, the default, is `--es-module-specifier-resolution=explicit`.) Package names are still just package names, e.g. `import fs from 'fs'`.
+
+This change isn’t because we’re trying to annoy an entirely new constituency now that `.mjs` is no longer required. We recognize that there are years’ worth of JavaScript out there written in the extensionless style, and this will feel like a huge breaking change for all that code. That’s why we’re providing the `--es-module-specifier-resolution=node` flag, and this is also subject to change before `--experimental-modules` itself ships unflagged.
```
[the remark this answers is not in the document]
weswigham:
```suggestion
We’re providing the `--es-module-specifier-resolution=node` flag to opt in to cjs-style extension and `index` resolution. We've turned it off by default to collect feedback on how users feel about using fully specified paths before we unflag the implementation. You can find our discussion on the topic [here](https://github.com/nodejs/modules/issues/268).
```

Especially don't mention "we didn't do this just to annoy you", because that's just going to rile up the people it annoys. I should know, I'm one of them, and reading it riled me. It's like the younger sibling saying "I'm not touching you" while holding their finger an inch away, pointed at you. It might technically be true, but the annoying effect is potentially magnified.

In addition, linking to our discussion on the topic lets those who are interested read the relevant background, and elides it for those who are not (although the issue doesn't collate all the meeting minutes on the topic, and there's a lot of them).

SMotaal:
@GeoffreyBooth reading over the [rendered](https://github.com/nodejs/modules/pull/291/files/e1e7563dced81a12214ebdabb088a56f8f6af5d8?short_path=a2b322d#diff-a2b322d311bab8f46d1ee729f17b2339) document, I have a more general consideration.

People will likely lose track of which `--experimental-modules` you will be referring to, not all share in the knowledge of more than one.

Does it make sense to always be explicit when first talking about either one in a separate context?

   Suggestions:

   |  *new* `--experimental-modules`  |  *old* `--experimental-modules`  | ❤️ |
   | :------------------------------: | :------------------------------: | - |
   |  `--experimental-modules` *v1*   |  `--experimental-modules` *v2*   | 👍|
   | `--experimental-modules` *alpha* | `--experimental-modules` *beta*  | 👎|

doc/announcement.md
robpalme: Agree with @weswigham - let's not mention tools.  Anything we state there is likely to become outdated.  It's also a catch-22 situation, because part of the reason for tools not updating is that they are waiting for a clear signal from Node.  Blog posts like *this* ought to be that clear signal rather than muddying the waters.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+### `--type` flag
+
+Use `--type=module`, or the shorthand `-m`, to run a `.js` file as an ES module.
+
+When running a file, e.g. `node file.js`, Node follows an algorithm to determine if it should load the file as CommonJS or as an ES module. First it looks for an explicit file extension (`.mjs` or `.cjs`); then it looks for a `"type"` field in the nearest parent `package.json`; and finally it looks at the `--type` flag. The `--type` flag can be `--type=module` (or `-m`) or `--type=commonjs`. You can also set this flag in `NODE_OPTIONS` if you’d like to change how Node behaves systemwide.
+
+This flag provides a way to support ES module syntax for input via `--eval`, `--print` or `STDIN`.
+
+## Mandatory file extensions
```
robpalme: This section title is a bit ominous until you read the detail.  It kinda also suggests that the underlying filenames themselves must have extensions, which is untrue.  Could we rename it to "Explicit filenames"?

doc/announcement.md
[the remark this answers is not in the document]
robpalme: Agree with @weswigham. The real danger line for me is "this will feel like a huge breaking change for all that code" - we shouldn't be referring to "breaking changes" in an experimental release and we shouldn't presume how people will feel.  The following paragraph already does the job of focusing on the positives/rationale, so I think we can just omit this paragraph.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+We’re making this change to align with browsers, where [extensions are required](https://github.com/WICG/import-maps#extension-less-imports). Part of the reason for Node.js to support ES modules in the first place is because ES modules are the standard, and the point of aligning to standards is to increase compatibility and interoperability between systems. In this case, Node.js supporting ES modules means that more JavaScript code will be capable of running in either Node or in browsers, without requiring some kind of translation process to make the code compatible with one or the other. Of course, it will always remain possible for users to write code that can’t function cross-platform (JavaScript referencing `window`, or importing `fs`, for example) but by aligning as much as we can, we encourage code to be cross-platform by default.
+
+Also, [import maps may bring support for automatic extension resolution to browsers](https://github.com/WICG/import-maps#extension-less-imports). It would be premature for Node to implement its own solution to supporting extensionless specifiers while efforts are underway to provide a cross-platform standard. Again, this is subject to change.
```
robpalme: Could we phrase this in a more collaborative way? It should be Node collaborating and tracking this proposal rather than Node being blocked by it.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+## `module.createRequireFromPath`
+
+The “CommonJS globals” `require`, `exports`, `module.exports`, `__filename`, `__dirname` are undefined in ES modules. However, [`module.createRequireFromPath()`](https://nodejs.org/api/modules.html#modules_module_createrequirefrompath_filename) can be used to create a CommonJS `require` function to be used in an ES module context.
```
robpalme: "undefined" -> "not defined"

doc/announcement.md
guybedford: It could also be worth mentioning "module" here, and why we may not be able to support it at all. So that those doubling down on this can be made aware.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+We’re making this change to align with browsers, where [extensions are required](https://github.com/WICG/import-maps#extension-less-imports). Part of the reason for Node.js to support ES modules in the first place is because ES modules are the standard, and the point of aligning to standards is to increase compatibility and interoperability between systems. In this case, Node.js supporting ES modules means that more JavaScript code will be capable of running in either Node or in browsers, without requiring some kind of translation process to make the code compatible with one or the other. Of course, it will always remain possible for users to write code that can’t function cross-platform (JavaScript referencing `window`, or importing `fs`, for example) but by aligning as much as we can, we encourage code to be cross-platform by default.
```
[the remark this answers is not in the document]
GeoffreyBooth: We should discuss this in the meeting. The point of requiring explicit extensions by default was to spark a discussion in the community, and part of that involves us explaining why that decision was made. I’m happy to tone this down if you think it’s too strong, but I don’t think we can just mention that filenames need extensions now with no context and no explanation.

When I mentioned not favoring any specific proposals, I meant for items not yet implemented.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+Currently it is not possible to create a package that is usable both via `require` (as CommonJS) and via `import` (as an ES module). There are efforts underway to address this, and may involve changes to the above (especially concerning `"main"`). Please do not publish any ES module packages intended for use by Node.js until this is resolved.
+
+## Coming soon
+
+All of the above is shipping as part of `--experimental-modules` in Node.js 12. On our [road map](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md) for improvements before the `--experimental-modules` flag is hopefully dropped in October 2019, when Node.js 12 reaches LTS status:
+
+- **Better loaders.** The current loaders code is a holdover from the previous `--experimental-modules` and doesn’t fully integrate with the new system. There is much work to be done here.
+
+- **Dual CommonJS/ES module packages.** We want to provide a standard way for package authors to publish a package that can be both `require`d into CommonJS or `import`ed into an ES module.
```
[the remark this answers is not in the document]
GeoffreyBooth: This section is just copied from our roadmap document.

The wording is careful to allow what you’re describing: an ESM-only package can still be “`require`d into CommonJS or `import`ed into an ES module” if `require` of ESM happens.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+- **Easier `require`.** Using `module.createRequireFromPath()` involves a lot of boilerplate. We hope to provide a simpler way to use `require` in ES module code.
+
+- **Package path maps.** Along the lines of the browser [import maps proposal](https://github.com/WICG/import-maps), we would like to support paths to files within packages. This would allow things like `import sdk from 'some-service/sdk'` to have `'some-service/sdk'` map to something like `./src/sdk/index.mjs`.
```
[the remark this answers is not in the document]
GeoffreyBooth: The proposal doesn’t mention bringing support to CommonJS, at least not yet. It hasn’t been updated in a while.

doc/announcement.md
GeoffreyBooth: I hear what you’re saying. I went with a version that tries not to have Node toot its own horn so much 😄 In this case, though, Node really is following.

doc/announcement.md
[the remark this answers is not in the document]
weswigham: I don't think if you need to convince people with extra learnings that it's a great idea... That it's that great of an idea.

I would much rather people gave feedback at this stage without being colored by our arguments for or against first, since that visceral reaction is what's going to spread, eg, on Twitter, upon the final release anyway~

And as I said, linking to the discussion via clickthru seems fine - the entire discussion is just too heavy for the announcement.

doc/announcement.md
SMotaal: Distinctions to be made here about publishing packages:

1. Not published and simply used locally for **non-production** testing — *highly encouraged*
2. Published to be used internally and strictly for **non-production** testing when warranted — *not recommended*
3. Published first time for public use as a third-party dependency — *discouraged*
4. Published as an update of the same — *highly discouraged*

Some people might need a little more to go on with respect to when packages are intended for Node.js and/or browser venues where they would want to ensure their ESM code is handled correctly making its way through those channels too. So I don't think "Please do not publish any ES module packages intended for use by Node.js until this is resolved." gives enough in way of what people should be experimenting with and what they should completely avoid, imho.

doc/announcement.md
ljharb: #293

GeoffreyBooth:
Let’s just settle this. The docs sub-group discussed and decided on the spelling “ES module” to match “CommonJS module”. If you agree with this, click 👍 below. If you prefer “ES Module” click 👎 below.

devsnek:
@geoffreybooth is that referring to a source text module or some abstract module?

GeoffreyBooth:
> @GeoffreyBooth is that referring to a source text module or some abstract module (which could be a stm)

Why do you ask, so you can point me to some spec text where everything is capitalized? This is a blog post, not spec text. An ES module is whatever the average user thinks it is, which I would think is probably a file or string containing `import` or `export` syntax, or a file or string intended to be loaded as if it had such syntax (i.e. an ambiguous file within an otherwise ES module project). “Loaded as an ES module” means that Node would evaluate the file expecting that it has or might have `import`/`export` syntax.

devsnek:
@GeoffreyBooth because if its some abstract module i'd say "es module" (i don't care about the capitalization there) and if its a source text module i'd say "source text module" (we already refer to it as such in the vm api)

GeoffreyBooth:
I think I’ve addressed all the feedback, at least that was clear on how I should address it. I’m not sure what changes to make regarding the dual packages/`"module"` paragraph, and @weswigham I made some changes to the extensions searching paragraph but I’m not sure all your concerns are addressed. Regarding that feature I think we need to explain our motivations (that was part of the discussion on flagging it) but I agree it should be presented as neutrally as possible, so if you want to suggest alternate language there we can continue to iterate.

doc/announcement.md
ljharb: I don't think this is resolved yet.

doc/announcement.md
ljharb: How is this something you can "now" do? This was something that was always planned; whether it happened to be shipped prior to the WG forming or not.

I do not want any implication that having an ESM `.js` file is suddenly an option because of this working group; we have merely preserved (and perhaps implemented) that option.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+For any file that Node.js tries to load, it will look for a `package.json` in that file’s folder, then that file’s parent folder and so on upwards until it reaches the root of the volume. This is similar to [how Babel searches for `.babelrc` files](https://babeljs.io/docs/en/config-files#file-relative-configuration). This new approach allows Node.js to use `package.json` for package-level metadata and configuration, similar to how it is already used by Babel and other tools.
+
+### `--type` flag
+
+Use `--type=module`, or the shorthand `-m`, to run a `.js` file as an ES module.
+
+When running a file, e.g. `node file.js`, Node follows an algorithm to determine if it should load the file as CommonJS or as an ES module. First it looks for an explicit file extension (`.mjs` or `.cjs`); then it looks for a `"type"` field in the nearest parent `package.json`; and finally it looks at the `--type` flag. The `--type` flag can be `--type=module` (or `-m`) or `--type=commonjs`. You can also set this flag in `NODE_OPTIONS` if you’d like to change how Node behaves systemwide.
+
+This flag provides a way to support ES module syntax for input via `--eval`, `--print` or `STDIN`.
```
ljharb:
```suggestion
This flag provides a way to support ES module syntax for input via `--eval`, `--print`, or `STDIN`.
```

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+We’re making this change to align with browsers, where [extensions are required](https://github.com/WICG/import-maps#extension-less-imports). Part of the reason for Node.js to support ES modules in the first place is because ES modules are the standard, and the point of aligning to standards is to increase compatibility and interoperability between systems. In this case, Node.js supporting ES modules means that more JavaScript code will be capable of running in either Node or in browsers, without requiring some kind of translation process to make the code compatible with one or the other. Of course, it will always remain possible for users to write code that can’t function cross-platform (JavaScript referencing `window`, or importing `fs`, for example) but by aligning as much as we can, we encourage code to be cross-platform by default.
```
ljharb: this is incorrect; extensions are **not** required in browsers - browsers ignore extensions entirely and have no opinion on them. Browsers require an exact URL, and any resolution or normalization (as in any other request on the web) needs to be handled by the server. This includes adding an extension, rewriting the path, adding or removing trailing slashes, etc.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+Also, [import maps may bring support for automatic extension resolution to browsers](https://github.com/WICG/import-maps#extension-less-imports). Should that proposal be accepted, Node is interested in supporting extensionless specifiers in a cross-platform compatible way.
```
ljharb: this is also inaccurate; it won't be "automatic" extension resolution, it will merely be clientside resolution. "adding an extension" is something that has to be statically precomputed by the server (as opposed to the server dynamically handling it at request time using rewriting rules)

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+## `module.createRequireFromPath`
+
+The “CommonJS globals” `require`, `exports`, `module.exports`, `__filename`, `__dirname` are not defined in ES modules. However, [`module.createRequireFromPath()`](https://nodejs.org/api/modules.html#modules_module_createrequirefrompath_filename) can be used to create a CommonJS `require` function to be used in an ES module context.
+
+## `import` for JavaScript only
+
+The previous `--experimental-modules` allowed `import` statements of JSON and native modules. This has been removed; you may use `module.createRequireFromPath()` for these.
+
+There is ongoing work in the standards space to cover JSON and WASM and other future potential module types. Node.js will add support for these in compliant ways when the standard is finalized.
```
ljharb: We should link to #255 and https://github.com/whatwg/html/issues/4315 here, and indicate that it's likely something we'll add in the future. There *is* no standard around importing JSON that applies to node, nor will there be - we'd be adding it unilaterally (with the intent of matching browsers) - so there's no finalization to wait for.

doc/announcement.md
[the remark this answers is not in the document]
ljharb: it would be better to discourage proliferation of patterns in the ecosystem prior to unflagging; it makes eventual migration to One True Pattern much harder for the community.

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Coming soon
+
+All of the above is shipping as part of `--experimental-modules` in Node.js 12. On our [road map](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md) for improvements before the `--experimental-modules` flag is hopefully dropped in October 2019, when Node.js 12 reaches LTS status:
+
+- **Better loaders.** The current loaders code is a holdover from the previous `--experimental-modules` and doesn’t fully integrate with the new system. There is much work to be done here.
+
+- **Dual CommonJS/ES module packages.** We want to provide a standard way for package authors to publish a package that can be both `require`d into CommonJS or `import`ed into an ES module.
+
+- **Easier `require`.** Using `module.createRequireFromPath()` involves a lot of boilerplate. We hope to provide a simpler way to use `require` in ES module code.
```
ljharb: I'm not sure this is necessary if `import` can be used for CJS. Can we indicate that?

doc/announcement.md
```diff
@@ -0,0 +1,99 @@
...
+
+There is ongoing work in the standards space to cover JSON and WASM and other future potential module types. Node.js will add support for these in compliant ways when the standard is finalized.
+
+## ES module code in packages
+
+_This is a work in progress and subject to change._ You can create a package with ES module sources by using the `package.json` `"main"` field to point to an ES module package entry point. Node will know to load it as an ES module if the file ends in `.mjs` or if the `package.json` also contains `"type": "module"`.
+
+Currently it is not possible to create a package that is usable both via `require` (as CommonJS) and via `import` (as an ES module). There are efforts underway to address this, and may involve changes to the above (especially concerning `"main"`). Please do not publish any ES module packages intended for use by Node.js until this is resolved.
+
+## Coming soon
```
ljharb:
```suggestion
## Possibly coming soon
```
given that not all these items have consensus, maybe best not to imply that they'll all actually happen

doc/announcement.md
GeoffreyBooth: The article is describing the new implementation as compared to the previous implementation. It’s not describing the new implementation’s _plan_ as compared to the possibly not fully implemented _plan_ for the previous implementation. I don’t think most people will know (or care) what the previous plan was, just what’s new in this version of the software versus the previous version.

doc/announcement.md
GeoffreyBooth: From a user’s perspective, extensions are required. They need to type `import './file.js'`, not `import './file'`. I think that getting into the weeds of MIME types and server resolution is too technical for a broad audience document.

doc/announcement.md
ljharb: Given the "we heard you" rhetoric at the beginning of the paragraph, this is absolutely implying that it wasn't part of the plan from day one, which it was. If you're talking about implementation differences and capabilities, then talking about "responding to feedback", especially flippantly implying that those of us involved prior to the WG a) weren't listening, or b) weren't planning on enabling this anyways, doesn't belong here.

doc/announcement.md
ljharb: `import './file'` will work fine if the server can serve that URL. I'm not talking about MIME types.

Note that rewrite rules to remove extensions for pages (altho usually not for assets) are absurdly common in PHP apache/httpd configs, and in Rails, so this isn't something that's likely to confuse most users.

GeoffreyBooth: > (altho usually not for assets)

Yes, and `.js` files are generally always assets. All the [examples](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Description) that I’ve ever seen of using `import` in a browser environment show the extension in the source code, so from the user’s perspective they need to type the extension.

We can change “extensions are required” to “extensions are (generally) required” to be more legalistically correct, but I think that just makes things all the more confusing as users will wonder when they might not be required, and I don’t think we should be discussing server-side rewrites here.

doc/announcement.md
GeoffreyBooth: This is just copied from the roadmap, so I wouldn’t make changes here that deviate from the roadmap.

To answer your direct point, though, as long as `require` can do things that `import` can’t (like include native modules, which we aren’t currently planning on supporting via `import`) then there will always be a need for `require` and `createRequireFromPath`, and so there will be a desire for getting `require` more easily. Maybe we never build it, but at the moment it’s on the roadmap.

doc/announcement.md
GeoffreyBooth: > this is absolutely implying that it wasn’t part of the plan from day one, which it was.

If supporting ESM in `.js` was ever the plan, shipping without any support for it certainly sent a message to the community that that wasn’t the final plan. I was an interested observer when the first `--experimental-modules` came out, probably a lot more interested than most, and I had no idea that ESM in `.js` was ever on Node’s roadmap.

Regardless, plans are never mentioned here. This text is just saying that we hear the request from the community loud and clear, and we’re responding to it. That’s not an indictment of anyone involved in the process so far. If anything, it shows that we’re good developers who are working hard to build the best product for our users.

GeoffreyBooth:
Okay, I’ve addressed all of the feedback posted above, either through the revisions in the document or through replies in the comments. Some of the comment threads didn’t have clear direction for me to know what to do about, as I’ve written above; if people can come to a consensus on what we want to do regarding those parts of the document, I’m happy to make those changes.

@MylesBorins [plans to open the PR for upstreaming on Saturday 2019-03-16](https://github.com/nodejs/ecmascript-modules/pull/60), and I think this should be merged in before then so that his PR can link to this document at its final URL in the repo. We can continue to revise it before it’s published to the world as our announcement blog post. I understand some folks still disagree with some of the things I’ve written here; I would ask that we merge this in and you can propose alternate text in PRs against this one, and those PRs can be debated, possibly in a meeting.

doc/announcement.md
ljharb: I don't think we can use the word "required" at all - browsers *do not care* about extensions.

doc/announcement.md
weswigham: > especially concerning `"main"`

That's just, like, your opinion maaaaaaan. But moreso: The extra call-out doesn't really add anything to the sentence.

ljharb:
I'd prefer no announcement whatsoever over an announcement that conveys a flippant tone about `.mjs` and "ESM in `.js` files"; I think it's better to leave this unmerged than to merge it before we're all on board with it.

doc/announcement.md
```diff
@@ -0,0 +1,101 @@
...
+
+We’re making this change to align with browsers, where [extensions are required](https://github.com/WICG/import-maps#extension-less-imports). Part of the reason for Node.js to support ES modules in the first place is because ES modules are the standard, and the point of aligning to standards is to increase compatibility and interoperability between systems. In this case, Node.js supporting ES modules means that more JavaScript code will be capable of running in either Node or in browsers, without requiring some kind of translation process to make the code compatible with one or the other. Of course, it will always remain possible for users to write code that can’t function cross-platform (JavaScript referencing `window`, or importing `fs`, for example) but by aligning as much as we can, we encourage code to be cross-platform by default.
+
+Also, [import maps may bring support for similar functionality to browsers](https://github.com/WICG/import-maps#extension-less-imports). Should that proposal be accepted, Node is interested in supporting extensionless specifiers in a cross-platform compatible way.
```
weswigham: We mention (and link to) import maps, like, four times in this post. Can we consolidate/reorganize all those bits to they're all together/unified (at least having the one call out three paragraphs below and then the one before this one and this one all seem slightly repetitive).

GeoffreyBooth: This is a fair point, but I’m not sure how to address it. The references are all very different:

- In the introduction, import maps are briefly mentioned as part of the context of ESM arriving in browsers.
- In the section on extension searching, they’re mentioned as a way to achieve “extensionless” specifiers in browsers.
- In the final “Works in progress” section, they’re mentioned as similar to our “package path maps” feature.

That’s it. I suppose we could remove the last mention entirely, though I feel it’s relevant; but the other two feel important to keep. How would you suggest changing things?

weswigham: Keep the sections, as is, remove the justifications and likening to import maps, put them into their own section after all the features have been introduced, maybe?

GeoffreyBooth:
> I’d prefer no announcement whatsoever over an announcement that conveys a flippant tone about `.mjs` and “ESM in `.js` files”

I removed the entire paragraph that was criticized for its tone. I don’t feel that the remaining [very brief mention](https://github.com/nodejs/modules/blob/announcement/doc/announcement.md#import-and-export-syntax-in-js-files) of `import` and `export` syntax in `.js` is flippant at all. Flippant is “not showing a serious or respectful attitude.” I don’t think there’s anything flippant about saying that we heard user feedback and addressed it; if anything, that _is_ respectful, of our community. I think telling users that we’re listening to them and responding to their feedback is critical to encouraging trust and adoption and making this entire effort a success. I think that’s the most important part of this document.

doc/announcement.md
ljharb: I'm still not OK with the wording in this paragraph. I'm not sure how to explain it any clearer, but please stop marking it as resolved; it is not.

weswigham: In terms of tone, I think it has much the same issue as the "we didn't do this to annoy you" line did. In isolation it might seem conciliatory, but mentioning it at all is only likely to incense anyone who feels these results are only a half-measure. It doesn't sound good to many of the people it's meant to ease the minds of.

doc/announcement.md
GeoffreyBooth: > Keep the sections, as is, remove the justifications and likening to import maps, put them into their own section after all the features have been introduced, maybe?

You mean don’t mention import maps at all in any of the three sections listed above? Why not? I think they really should be mentioned at least in the intro paragraph about ESM in browsers.

Import maps isn’t our proposal and we’re not implementing any of it in Node, so I feel like it would be quite odd for it to get its own section.

I can see import maps as not being essential to mention in the two feature sections, if you think those mentions detract. Would you still want those mentions removed even if import maps doesn’t get its own section?

weswigham: We won't actually be _implementing_ import maps though - in no way do they indicate a feature on their own. They are merely some work we're keeping in mind as we work. We'd be free to come to the same feature set without import maps existing if we so desired. I just think that if you think the justifications are for some reason warranted, rather than repeating it three times, doing it all at once in a single section at the bottom would scan much better.

doc/announcement.md
robpalme: All we need is a simple statement that modules can use the JS extension.

That will be enough. It's super clear. 

We should avoid referencing individuals that have taken particular positions in the past, because that frames it as a conflict between different sides, whereas it's more of a collaboration.

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
+# Announcing a New `--experimental-modules`
+
+Back in 2017, Node.js 8.9.0 shipped experimental support for [ECMAScript modules](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/), known for their `import` and `export` statements. This support was behind the flag `--experimental-modules`.
+
+A lot has happened since then. All major browsers [now support](https://caniuse.com/#feat=es6-module) ECMAScript modules (ES modules) via `<script type="module">`. Various projects have sprung up to make npm packages with ES module sources available for use in browsers via `<script type="module">`. Support for [import maps](https://github.com/WICG/import-maps), which bring to browsers Node-style package names in `import` statements, [is coming to Chrome](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/qYeQFqgFOyA/rXJapjMaEAAJ).
+
+All this progress toward adoption of ES modules has increased the urgency for Node.js to ship its support for ES modules. Now that there are other runtimes and environments where ES modules are in use, it’s more important than ever that Node.js support the JavaScript standard.
```
guybedford: nit - perhaps "**this** JavaScript standard" to be clear?

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+Node’s initial ES modules support remained in an experimental state in order to allow the community time to provide feedback on that design. The [Modules Team](https://github.com/nodejs/modules) was formed to act on this feedback and ship first-class support for ES modules in Node.js. This has led to a [new implementation](https://github.com/nodejs/ecmascript-modules) for supporting ES modules, that we are pleased to announce will ship as part of Node.js 12. It will replace the old `--experimental-modules` implementation, behind the same flag. We hope that this new implementation addresses many of the community’s concerns, and can ship as part of Node.js proper, without a flag, before Node.js 12 reaches [LTS status](https://nodejs.org/en/about/releases/) in October 2019.
```
guybedford: Node's -> Node.js's. I believe the policy is never to use "Node" without the ".js".

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Files ending in `.mjs` are explicitly treated as ES modules in `import` statements and when run via the `node` command.
+
+And the new version of `--experimental-modules` adds:
+
+## `import` and `export` syntax in `.js` files
+
+We heard some [very](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) [strong](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054) [feedback](https://twitter.com/maybekatz/status/1062473765865512961) that Node needs to provide a way to use `import` and `export` syntax in `.js` files.
+
+The new `--experimental-modules` provides two ways: a `"type"` field in `package.json` and a `--type` flag. Their naming was inspired by browsers’ `<script type="module">`.
```
guybedford:
```suggestion
The new `--experimental-modules` provides two ways to achieve this: a `"type"` field in `package.json` and a `--type` flag for the entry point. Their naming was inspired by browsers’ `<script type="module">`.
```

Perhaps also add:

```
Ideally setting this flag can become part of the package manager init process.
```

or something to that affect?

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+Add `"type": "module"` to the `package.json` for your project, and Node will treat all `.js` files in your project as ES modules.
+
+If some of your project’s files use CommonJS and you can’t convert your entire project all at once, you can either rename those files to use the `.cjs` extension or put them in a subfolder containing a `package.json` with `{ "type": "commonjs" }`, under which all `.js` files are treated as CommonJS.
+
+For any file that Node.js tries to load, it will look for a `package.json` in that file’s folder, then that file’s parent folder and so on upwards until it reaches the root of the volume. This is similar to [how Babel searches for `.babelrc` files](https://babeljs.io/docs/en/config-files#file-relative-configuration). This new approach allows Node.js to use `package.json` for package-level metadata and configuration, similar to how it is already used by Babel and other tools.
+
+### `--type` flag
+
+Use `--type=module`, or the shorthand `-m`, to run a `.js` file as an ES module.
```
guybedford: Perhaps be clear about the full usage here in the example - `node --experimental-modules --type=module main.js`?

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+A separate flag `--experimental-json-modules` enables experimental support for `import` of JSON files. There is [work in progress](https://github.com/whatwg/html/issues/4315) for standardizing this feature with browsers, and Node hopes to align our support with the eventual standard.
+
+There is also ongoing work to cover WASM and other future potential module types. Node.js will add support for these in spec-compliant ways over time.
+
+## ES module code in packages
+
+_This is a work in progress and subject to change._ You can create a package with ES module sources by using the `package.json` `"main"` field to point to an ES module package entry point. Node will know to load it as an ES module if the file ends in `.mjs` or if the `package.json` also contains `"type": "module"`.
+
+Currently it is not possible to create a package that is usable both via `require` (as CommonJS) and via `import` (as an ES module). There are efforts underway to address this, and may involve changes to the above (especially concerning `"main"`). Please do not publish any ES module packages intended for use by Node.js until this is resolved.
```
guybedford: I'm not sure how genuine this paragraph is.

Perhaps be more explicit - "it is not possible to have a package main that is usable by both `require` and `import`"?

Even that isn't quite right as CommonJS is perfectly usable by import.

Perhaps the exact version is "it is not currently possible to have a package main that exposes an ES module for `import` and a CommonJS module for `require`"?

Then it would be worth using the term "dual mode" here as something that is being explored.

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Works in progress
+
+All of the above is shipping as part of `--experimental-modules` in Node.js 12. On our [road map](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md) for potential improvements before the `--experimental-modules` flag is hopefully dropped in October 2019, when Node.js 12 reaches LTS status:
+
+- **Better loaders.** The current loaders code is a holdover from the previous `--experimental-modules` and doesn’t fully integrate with the new system. There is much work to be done here.
```
guybedford: Can we rather have:

```
**Loaders features.** Loaders are being further developed to support process isolation, multiple loaders, multi-Realm support with lower-level hooks. The `--loader` API will still change considerably before this is unflagged.
```

or something along those lines.

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+- **Dual CommonJS/ES module packages.** We want to provide a standard way for package authors to publish a package that can be both `require`d into CommonJS or `import`ed into an ES module.
```
guybedford: Perhaps `We're investigating ways for` instead of "We want to provide"? That is, this something we're looking into as discussed previously.

doc/announcement.md
GeoffreyBooth: > Perhaps the exact version is “it is not currently possible to have a package main that exposes an ES module for `import` and a CommonJS module for `require`“?

How is this meaningfully different than “Currently it is not possible to create a package that is usable both via `require` (as CommonJS) and via `import` (as an ES module).”? I didn’t want to mention “main” specifically here in case the ESM entry point ends up in a different field.

doc/announcement.md
GeoffreyBooth: I thought that only applied to documentation. Does it apply to all official communications?

doc/announcement.md
GeoffreyBooth: A _lot_ of back-and-forth went into this section, I don’t want to upset any compromises by changing it further if you don’t mind.

doc/announcement.md
GeoffreyBooth: I think our discussions are pretty firm in the camp of us wanting to make this happen one way or another, and it’s on our roadmap. Several people expressed concerns about fracturing the npm registry if we don’t provide a way for dual packages to happen. I think we can be as forthright as “We want to provide”.

doc/announcement.md
guybedford: Well you can create a package that exposes both CommonJS and ES Modules just fine through `require('pkg/main.js')` and `import 'pkg/main.mjs'`. So this is specifically a main thing. I don't think there is any confusion in mentioning main by name, and not doing that makes this sound like we are now the package police.

doc/announcement.md
guybedford: The changes are purely stylistic edits, as opposed to content.

doc/announcement.md
guybedford: Then `We would like to provide` perhaps rather? I'm just hesitant to make future promises.

doc/announcement.md
GeoffreyBooth: I want to join the package police. Where do I sign up?

I think I found a phrasing that avoids the issue that I found confusing where you mean “main” to be “package entry point” but I immediately think of it as meaning the `"main"` field (which defines the same thing, of course, but potentially only for CommonJS). How about this:

> Currently it is not possible to create a package that can be used via both `require('pkg')` and `import 'pkg'`.

That avoids needing the word “main” and it makes clear we’re talking about the bare specifier/package entry point.

guybedford: > I want to join the package police. Where do I sign up?

You already did :)

> Please do not publish any ES module packages intended for use by Node.js until this is resolved.

Just to be really clear on this sentence, because I'm not sure what it means for me as a library author, are we effectively saying: "Be aware that publishing packages that rely on the "main" being a module should not be published because the exact semantics of ES modules are still changing and thus that package could break in future"?

I think it would help to be really really clear on what the line is for users here.

To try and spell it out, here is my best guess:

It is NOT ok to:

* Publish packages with "type": "module"
* Publish packages with "main": "main.mjs" being a module

as the above may break as the semantics change under --experimental-modules.

But it IS ok to:

* Publish packages that contain ES modules for consumption by build tools
* Publish packages with "module": "main.js" as this will likely not be supported
* Publish packages with require('esm')

Does that cover what you mean? It would help to understand.

GeoffreyBooth: That covers what I mean, though I'm not warning against `type: "module"` because that's not on our roadmap as something lacking consensus or that might change in phase 3. But everything else, yes.

If you want to draft alternative language I'm happy to change this. One note I got from Wes though was to keep it short, so I guess somehow get all that across without the two bulleted lists?

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Files loaded as ES modules are loaded in strict mode, which in CommonJS requires adding `'use strict';` to the top of every file.
+
+- Files ending in `.mjs` are explicitly treated as ES modules in `import` statements and when run via the `node` command.
+
+And the new version of `--experimental-modules` adds:
+
+## `import` and `export` syntax in `.js` files
+
+We heard some [very](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) [strong](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054) [feedback](https://twitter.com/maybekatz/status/1062473765865512961) that Node.js needs to provide a way to use `import` and `export` syntax in `.js` files.
```
ljharb: Thanks, this is much better - I'd still prefer to indicate that this was always part of the plan (just was never planned to be the default, as it still is not).

GeoffreyBooth: This language was the result of a lot of back and forth between various people, so I'd rather not open it up again for debate if you don't mind? Could this be acceptable to you?

ljharb: I'm not sure why "how many people that weren't me" you talked to should mean that I don't get to retain my opinion here.

GeoffreyBooth: Looking at the previous implementation’s [user-facing docs](https://nodejs.org/api/esm.html) and [design doc](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md), the previous plan for supporting ESM in `.js` files was via a loader as shown in the example [here](https://nodejs.org/api/esm.html#esm_resolve_hook).

How about we change “Node.js needs to provide a way to use `import` and `export` syntax in `.js` files” to “Node.js needs to provide a way to use `import` and `export` syntax in `.js` files without requiring a loader.” Would that be acceptable?

cc @weswigham @robpalme

bmeck: Guy and myself had plenty of talks about different ways to declare formats prior to the while forming, if you need me to dig up gists etc. I can, but the talks did occur both in EP and after the initial implementation. I am rather busy but is it really necessary to bicker about the timelines so much?

robpalme: My original point was that the history doesn't matter. 

Whilst I'm happy with the wording as-is, if you wish to discard the historical/controversy angle, that's fine too. The content already sells itself, so the more boring/factual the better IMO.

ljharb: per offline discussion:
```suggestion
We heard feedback that Node.js needs to provide a way to use `import` and `export` syntax in `.js` files.
```

SMotaal: It may be worthwhile to note that feedback is different from what was heard, imho wording this needs to balance from both sides and be slightly biased towards the reader (ie they will be doing the effort to read and use this after all).

```suggestion
We are make a lot of progress towards popular features demanded by members of the community and in ways that will ensure that all existing projects can continue to operate smoothly.

With this new implementation, we are making it possible for the first time to actually `import` and `export` syntax in `.js` files with opt-in package configuration or CLI flags directly, ie no longer requiring a custom loader.
```

This suggestion focuses on the substance, **but** I sincerely think that references from which we claim to infer such demand are necessary somewhere in the announcement — it is appropriate blogging decorum and ensure tracking with a proper chronological structure.

So we need a way to relate to those links:

- https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md
- https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054
- https://twitter.com/maybekatz/status/1062473765865512961

I suggest considering a second statement in the first paragraph making the case from those links that is addressed in the next paragraph. This would flow well for the reader and be very balanced to both Node.js and the community.

I worry that being too sensitive here comes off negatively for Node.js — did we not just give a solution that was asked for here after insisting we will never ever do that?! If we tiptoe about that, the reader will be inclined to think it, but if we assert the correct narrative, it keeps readers focused on the exciting new features, and frames those links as one community working towards one goal.

SMotaal: That suggestion (which I strongly recommend would look more like this):

```suggestion
We are make a lot of progress towards some [popular features demanded by members of the community](https://twitter.com/maybekatz/status/1062473765865512961) and in ways that will ensure that all existing projects can continue to operate smoothly. This required a lot of careful consideration and insights from community discussions on how [the community uses ESM in projects today](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) and  [possible ways for packages to move forward](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054).

With this new implementation, we are making it possible for the first time to actually `import` and `export` from `.js` files with opt-in package configuration or CLI flags directly, ie no longer requiring a custom loader.
```


<details><summary>Preview</summary><blockquote>

We are make a lot of progress towards some [popular features demanded by members of the community](https://twitter.com/maybekatz/status/1062473765865512961) and in ways that will ensure that all existing projects can continue to operate smoothly. This required a lot of careful consideration and insights from community discussions on how [the community uses ESM in projects today](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) and  [possible ways for packages to move forward](https://gist.github.com/ceejbot/b49f8789b2ab6b09548ccb72813a1054).

With this new implementation, we are making it possible for the first time to actually `import` and `export` from `.js` files with opt-in package configuration or CLI flags directly, ie no longer requiring a custom loader.

</blockquote></details>

ljharb: No, because nobody insisted we would never do that. With or without any of this feedback, we would always have provided some way to have ESM in .js - predating the WG as well. Thus, the links aren’t relevant, and would further the misconception that this is a newly intended feature.

MylesBorins:
Would a better path forward be to remove any language that has any sort of
bias... Avoid history... Talk about facts of new implementation only

ljharb:
I think so - I think [this language](https://github.com/nodejs/modules/pull/291#discussion_r268399754) geoffrey proposed in offline discussion achieves that.

ljharb:
@GeoffreyBooth and I discussed things a bit offline; if by tomorrow night the two of us haven't been able to agree on better text, I'm going to go ahead and stamp with the current phrasing (also OK with [this tweak](https://github.com/nodejs/modules/pull/291#discussion_r268012749), but not required).

doc/announcement.md
```diff
@@ -0,0 +1,103 @@
+# Announcing a New `--experimental-modules`
```
zackschuster: nit: the `New` here should be `new` (all lowercase)

doc/announcement.md
SMotaal: @ljharb can you clarify what the intent was:

1. Allow a `package.json` to make it possible to `node main.js` to be ESM: Yes/Loader/No
2. Allow a flag `node --flag main.js` to be ESM: Yes/Loader/No
3. Allow a package to be published with `.js` files as ESM: Yes/Loader/No
4. Allow a package to be published with `.js` files as ESM & CJS: Yes/Loader/No

Can you reply on each one to help me get a more complete picture please.

Because from the end user's side (which I was at roughly 12 months ago) things looked a certain way, and regardless of what the intent was, it is important to recognize that no matter how hard we all try, the other perspective is always very hard fully appreciate — and naturally mine was one of a the wide spectrum of perspectives, but not an outlier

**As such**, it is important for us to be as objective and thorough as possible when talking to the audience.

ljharb: @SMotaal 
1. afaik undiscussed, but i don't see any reason why we wouldn't have - if package.json has any control then it should have entry point control.
1. yes, since day one, if for nothing else for stdin/eval/extensionless files.
1. an open question, but leaning towards yes because if it can work for non-bare imports via package.json (ie, for your own top-level files) than it wouldn't make sense to actively prohibit it for published files.
1. dual packages have always been intended; without that as an option, there's no migration strategy.

It is totally valid that some user perspectives may have been solely informed from the implementation; it's also valid that user perspectives *do not get to dictate* what the intention was.

In general, being objective also means not validating "internet mob outrage made change happen" so as to ensure that's not seen as an effective means to effect change in the future.

GeoffreyBooth:
In all seriousness, I have heartfelt appreciation for all the careful feedback that everyone has given this, and I’m very encouraged that we were able to reach consensus. _Thank you_ to everyone who took the time to comment or review.

That said, I’m gonna merge before anyone else can object 😉This isn’t truly final until it’s posted on Medium, which I assume won’t happen until after Node 12 ships, so there’s still time for more revisions via PR between now and then.

GeoffreyBooth:
P.S. Just realized I shouldn’t delete the branch while there still might be links to it, such as in our upstream PR. If people could update wherever they might’ve posted the link to use the new URL, that would be great:

https://github.com/nodejs/modules/blob/master/doc/announcement.md

doc/announcement.md
SMotaal: Thanks for clarifying

PR #304, opened 2019-03-28
Update documents per current status

GeoffreyBooth:
There shouldn’t be anything controversial in here, this just updates the various documents in this repo to reflect that the fact that our upstream PR has been merged (🎉) and we’re now in Stage 3; and `--type` became `--entry-type`.


doc/announcement.md
```diff
@@ -48,13 +48,15 @@ If some of your project’s files use CommonJS and you can’t convert your enti
...
+When running a file, e.g. `node --experimental-modules --entry-type=module main.js`, Node.js follows an algorithm to determine if it should load the file as CommonJS or as an ES module. First it looks for an explicit file extension (`.mjs` or `.cjs`); then it looks for a `"type"` field in the nearest parent `package.json`; and finally it looks at the `--entry-type` flag. The `--entry-type` flag can be `--entry-type=module` or `--entry-type=commonjs`.
 
-This flag provides a way to support ES module syntax for input via `--eval`, `--print`, or `STDIN`.
+Note that setting the type of a file via `--entry-type` does _not_ set the type for any files that that file may `import`; you probably want to add a `package.json` with a `"type"` field to define the type for a folder of files.
+
+This flag provides a way to support ES module syntax for input via `--eval`, `--print`, or `STDIN`; and for loose single `.js` files outside of any package or project.
```
ljharb:
```suggestion
This flag provides a way to support ES module syntax for input via `--eval`, `--print`, or `STDIN`; or for an extensionless entry point; and for loose single `.js` files outside of any package or project.
```

GeoffreyBooth: It would still need to be outside a package scope though, since `"type"` should govern extensionless files. How about “loose single .js or extensionless file”?

ljharb: Sure, that works - altho i thought entry-type overrides package.json “type” for the entry point? or does it throw in that case.

PR #343, opened 2019-06-19
Updates per 2019-06-19 meeting

GeoffreyBooth:
Look I took notes! (Sorta.)

### DIFF
--- a/doc/plan-for-new-modules-implementation.md
+++ b/doc/plan-for-new-modules-implementation.md
@@ -99,24 +99,23 @@ Phase 3 improves user experience and extends the MVP. Phase 3 is malleable based
   - Should loaders be per package, per application or either?
   - **Status**: In development.
 
-* Map the paths within modules, providing similar functionality as the browser’s [import maps proposal](https://github.com/WICG/import-maps#packages-via-trailing-slashes).
+* `"exports"` field: for consumers of a package, map the paths of deep imports to provide encapsulation (an explicit public API); pretty specifiers (no file exensions or paths that include things like `dist/`) and flexibility for future package versions renaming or moving files without affecting the package’s public API. Applies to both ESM and CommonJS.
+  - Proposal: [Package Exports Proposal](https://github.com/jkrems/proposal-pkg-exports).
+  - **Status**: Ready for PR.
+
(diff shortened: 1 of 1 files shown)


doc/plan-for-new-modules-implementation.md
```diff
@@ -99,25 +99,24 @@ Phase 3 improves user experience and extends the MVP. Phase 3 is malleable based
   - Should loaders be per package, per application or either?
   - **Status**: In development.
 
-* Map the paths within modules, providing similar functionality as the browser’s [import maps proposal](https://github.com/WICG/import-maps#packages-via-trailing-slashes).
+* `"exports"` field: for consumers of a package, map the paths of deep imports to provide encapsulation (an explicit public API); pretty specifiers (no file exensions or paths that include things like `dist/`) and flexibility for future package versions renaming or moving files without affecting the package’s public API. Applies to both ESM and CommonJS.
+  - Proposal: [Package Exports Proposal](https://github.com/jkrems/proposal-pkg-exports).
+  - **Status**: Ready for PR.
+
+* `"imports"` field: provide mappings for `import` specifiers within a package, similar to browsers’ [import maps proposal](https://github.com/WICG/import-maps#packages-via-trailing-slashes).
```
MylesBorins: this is not accurate per the conversation in the meeting. I think we discussed exports as being something we had consensus on moving for with whereas imports was something to be further discussed

GeoffreyBooth: Yes, and isn’t that what I wrote?

MylesBorins: To be very clear, I think it is premature to include this proposal at this time

GeoffreyBooth: I can remove `"imports"`, I guess that’s what you’re saying? Though I didn’t think we had such a high bar for adding things to this doc? We already have a status saying that it’s nowhere close to consensus.

MylesBorins: We have had a high bar for removing things from the doc and the doc has been pointed to as a source of truth regarding things we are exploring. I would prefer we don't add it for now

PR #392, opened 2019-09-30
doc: add missing sept notes

MylesBorins:
Refs: https://github.com/nodejs/modules/issues/383
Refs: https://github.com/nodejs/modules/issues/386

### DIFF
--- a/doc/meetings/2019-09-11.md
+++ b/doc/meetings/2019-09-11.md
@@ -1,0 +1,96 @@
+# Node.js Foundation Modules Team Meeting 2019-09-11
+
+## Links
+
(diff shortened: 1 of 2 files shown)


guybedford:
This has taken a while to land - do we still require quorum from the meeting to land these notes? If so we should ensure to include in the next agenda. Also we should probably revert https://github.com/nodejs/modules/pull/398 as well if I've been to hasty there.

GeoffreyBooth:
Docs/notes don’t require quorum, they just need to be open PRs for at least three days without anyone requesting changes.

guybedford:
Thanks @GeoffreyBooth, merging this then.

PR #397, opened 2019-10-07
doc: move fishrock123 to observer

Fishrock123:
I'm not going to actively participate in my free time.


guybedford:
Thanks @Fishrock123 for our participation in the process to date. I'm assuming we don't need quorum to land this since this is entirely your decision, so will go ahead and merge.

PR #429, opened 2019-11-14
Link to dropping the flag 🎉

dandv:


GeoffreyBooth:
Thanks for this! I think we probably want to update this when the unflagged version is released, probably next week in the next v13 minor.

MylesBorins:
Closing this as a dupe since the info has already landed in the repo afaict

thanks for the PR @dandv

Issue #473, opened 2020-01-17
Rewrite modules readme


GeoffreyBooth:
I just listened to https://syntax.fm/show/211/hasty-treat-modules-in-node and the speakers there seemed to think that named exports for CommonJS were coming soon. They got this idea from the [readme in this repo](https://github.com/nodejs/modules/blob/master/README.md), because it still lists every feature we ever considered implementing, including “Named exports when importing CJS (#81)”. When that was written, the purpose of the README was to be a document setting down our agreed-upon consensus.

Now that ESM in Node has shipped, we don’t really need our consensus spelled out anymore. I think people coming to this repo today are likely to be looking for information about what’s on the roadmap for future improvements to ES modules in Node. I think the readme should answer that question directly, both in terms of what we’re working on or would be open to contributions for solving; and what _isn’t_ likely to come to Node, with reasons why. This could also be how we retire the roadmap, moving its purpose into the readme (but the readme doesn’t need to include details about past phases; we could keep all that in some separate document).

Thoughts? Anyone like to collaborate on this?

weswigham:
I would _hope_ that spurs us to reinvest in making that actually happen, once the core's shipped? Most of the reason that hasn't really advanced is procedural, IMO (and we've been doing more productive things in the meantime). We do still have philosophical buyin form the tc39 to make named exports happen AFAIK, it's just going to be a long process to work through it with the committee.

bmeck:
@weswigham to my knowledge TC39 wanted a different approach from the Dynamic Modules proposal which would affect existing module graphs so idk where/what we could invest in currently. I'm listed as the champion but was pondering closing that proposal as I don't see a path forward for that one at least.

GeoffreyBooth:
We could also just explain the situation in more detail. It doesn't have to be go/no go, we could say we want to do it but can't find a way currently, and here's why.

weswigham:
@bmeck IIRC you were essentially told that they didn't wanna deal with it in the SourceTextModuleRecord spec and that they should look into removing SourceTextModuleRecord from the main spec, which essentially shut you down, since that's not what you were looking for. The way forward, IMO, is to make that happen and bring that forward, as they proposed. (Where it probably gets shut down by other people and we can return to productively discussing how to actually integrate it into the spec). It's roundabout and disheartening, but probably the process that needs to happen, from what I understand.

bmeck:
[...]

PR #482, opened 2020-02-09
Rewrite readme

GeoffreyBooth:
Resolves https://github.com/nodejs/modules/issues/473. This archives out-of-date content and puts on our README the most important information that most visitors to this repo are likely looking for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 ## Purpose
 
-The Node.js Modules Team maintains and actively develops the ECMAScript Modules (ESM) implementation in Node.js Core.
+The Node.js Modules Team maintains and actively develops the ECMAScript Modules (ESM) implementation in Node.js core.
 
 Work includes:
 
-* ESM Loader implementation
-* ESM Module Specifier Resolution Algorithm
-* Loader Hooks
-* ESM and CommonJS Interoperability
-* Node.js and Browser interoperability
-* VM Modules implementation
+* ESM loader implementation
+* ESM specifier resolution
+* Loader hooks
+* ESM and CommonJS interoperability
+* Node.js and browser interoperability
+* VM modules implementation
 
 ## Current Efforts
 
--- a/README.md
+++ b/README.md
@@ -19,59 +19,6 @@ Work includes:
 
 A new ECMAScript modules implementation shipped behind the `--experimental-modules` flag in Node.js 12.0.0. The flag was dropped in Node.js 13.2.0, although the implementation remains experimental and is still subject to change. A road map of the implementation’s development is [here](./doc/plan-for-new-modules-implementation.md).
 
-
-## Features
-
-Based on [these use cases](./doc/use-cases.md) ([#55](https://github.com/nodejs/modules/issues/55)), our implementation aims to support the following features (subject to change):
-
-### Baseline Modules Implementation Features:
-* Spec compliance ([#132](https://github.com/nodejs/modules/issues/132))
-* Browser equivalence ([#133](https://github.com/nodejs/modules/issues/133))
-* Don’t break CommonJS ([#112](https://github.com/nodejs/modules/issues/112))
-* No refactoring ([#87](https://github.com/nodejs/modules/issues/87))
-
-#### Browser Interop:
-* Browser and Node compatibility without building ([#107](https://github.com/nodejs/modules/issues/107))
-* Browser-compatible specifier resolution ([#109](https://github.com/nodejs/modules/issues/109))
-* Browser-compatible builds for ESM and CommonJS ([#108](https://github.com/nodejs/modules/issues/108))
-
-#### CommonJS Interop:
-* Named exports when importing CJS ([#81](https://github.com/nodejs/modules/issues/81))
-* Multi-mode packages ([#94](https://github.com/nodejs/modules/issues/94))
-* Transparent interoperability for ESM importing CommonJS ([#100](https://github.com/nodejs/modules/issues/100))
-* Consumer-agnostic imports ([#105](https://github.com/nodejs/modules/issues/105))
-* Mixed module types within app/module; gradual migration from CommonJS to ESM ([#99](https://github.com/nodejs/modules/issues/99))
-* ESM in .js files ([#151](https://github.com/nodejs/modules/issues/151))
-
-### Existing Node.js Utility Features:
-
-* Importing non-JavaScript files ([#115](https://github.com/nodejs/modules/issues/115))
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,75 +2,64 @@
...
-* Browser equivalence ([#133](https://github.com/nodejs/modules/issues/133))
-* Don’t break CommonJS ([#112](https://github.com/nodejs/modules/issues/112))
-* No refactoring ([#87](https://github.com/nodejs/modules/issues/87))
+	CommonJS is not statically analyzable (that’s one of the primary benefits of `import`/`export` syntax) and therefore the exported names cannot be known to Node.js simply by parsing JavaScript source code: the code needs to actually be evaluated in order for Node.js to discover what identifiers get attached to `module.exports` for each file. But evaluating the CommonJS code in what the ECMAScript Modules spec calls the “linking” phase (when the module graph is determined) is not permitted; this would mean that CommonJS code is evaluated early, before any ES module code, whereas the spec defines the order of modules’ evaluation based on the depth of imports. This early evaluation presents other issues in that if a dependency gets converted from CommonJS to ESM, its order of evaluation would change in the user’s application; and early evaluation might trigger side effects if reading the values from `module.exports` were getters or proxies.
```
guybedford: > and early evaluation might trigger side effects if reading the values from `module.exports` were getters or proxies.

This is a separate point that isn't about early evaluation. This is specifically about `Object.value(module.exports)` having side effects, whereas those side effects would _never_ otherwise trigger.

README.md
```diff
@@ -2,75 +2,64 @@
...
-* Node.js and Browser interoperability
-* VM Modules implementation
+* ESM loader implementation
+* ESM specifier resolution
+* Loader hooks
+* ESM and CommonJS interoperability
+* Node.js and browser interoperability
+* VM modules implementation
 
 ## Current Efforts
```
guybedford: It would be nice if we could include a `Current Status` section before all the experimental / new stuff, that captures what the group has already achieved.

GeoffreyBooth:
> The overall content is great, I'm just thinking perhaps we need to separate the README from the `history.md` from the `roadmap.md`. We also don't want to risk the README content going stale again too.

Wherever we put the “this is what we’re working on now” content, it will gradually become stale. I think it belongs near the top of the README, because that’s what the general public is coming to our repo for. That’s the information they can’t get anywhere else. If they want to know what we’ve accomplished, they can read the ESM docs.

> If we're looking to make specific advisories to users about the implementation status, perhaps the better place for that is the modules docs?

We do this as well; see the [loaders](https://nodejs.org/api/esm.html#esm_experimental_loaders) section with its plentiful warnings.

README.md
GeoffreyBooth: Can you please rephrase the text so that it’s correct? Maybe offer a suggestion?

README.md
DerekNonGeneric: It sounds like a good idea, I'm just a bit unclear on what you have in mind. What would the different statuses be? [Here's an idea](https://github.com/DerekNonGeneric/loader339#status-per-goal) I had for displaying different statuses of various features, it might be similar to what you're after.

DerekNonGeneric:
Please take notice that a link, which I've emboldened below, exists in the following sentence of the [introduction of the ECMAScript Modules docs page](https://nodejs.org/api/esm.html#esm_introduction) that currently expects [`doc/plan-for-new-modules-implementation.md`](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md) to exist in its current location.

> Node.js contains support for ES Modules based upon the Node.js EP for ES Modules and the **ECMAScript-modules implementation**.

This PR, if merged, would move it to the archives causing that URL to turn up nothing. I recommend we keep a page at `doc/plan-for-new-modules-implementation.md` (maybe for a few months) informing people that the page has moved and link to its new home in the archives.

README.md
```diff
@@ -2,75 +2,64 @@
...
 
-> ### [Current Plan for ECMAScript Modules support in Node.js](./doc/plan-for-new-modules-implementation.md)
+### To Do / In Progress
```
guybedford: To be very blunt then, since you have asked me to be, I simply do not think this content is suitable for our readme. Since it is rapidly changing, likely to get out of date, and not the right information for someone "interested in the Node.js modules group".

As said before, I think we should separate this out into `roadmap.md` and `previous-work.md`.

README.md
guybedford:
```suggestion
	CommonJS is not statically analyzable (that’s one of the primary benefits of `import`/`export` syntax) and therefore the exported names cannot be known to Node.js simply by parsing JavaScript source code: the code needs to actually be evaluated in order for Node.js to discover what identifiers get attached to `module.exports` for each file. But evaluating the CommonJS code in what the ECMAScript Modules spec calls the “linking” phase (when the module graph is determined) is not permitted; this would mean that CommonJS code is evaluated early, before any ES module code, whereas the spec defines the order of modules’ evaluation based on the depth of imports. This early evaluation presents other issues in that if a dependency gets converted from CommonJS to ESM, its order of evaluation would change in the user’s application.
	
	 In addition, even if these issues were surmountable, the way ES modules define their exports still means we would be applying an `Object.values(module.exports)` for all CommonJS modules as soon as they finish executing, in order to populate the ES module exports. This can lead to a number of problems:
	 
	  * The triggering of getters and proxies (eg for dynamic generation / deprecation warnings) can throw errors or cause unwanted side effects.
	  *  New exports added to `module.exports` after this time would never be available as named exports.
	  * Mutations to `module.exports` cannot be kept live with the ES module exports, unless we turn all CommonJS `module.exports` into a proxy. This in turn could have backwards compatibility issues.
```

README.md
DerekNonGeneric: Latest commit in my PR to this branch may be what you're after.

https://github.com/nodejs/modules/pull/487/commits/eabe9c231e83538048018c32391a8034040e89ef

jkrems: > As said before, I think we should separate this out into roadmap.md and previous-work.md.

Would it be better to have something like a project board for the roadmap? So it can properly link to issues/PRs, including automatic status updates as things get closed?

DerekNonGeneric: @jkrems, the argument against that approach, which was brought up to me when I proposed the exact same idea to @GeoffreyBooth, is that these things need to be archivable and a more-static public record of them needs to be kept.

GeoffreyBooth:
@guybedford please review. This version just archives the features and old road map, without adding any new content. This will satisfy #473 while we figure out the best place to put that other content (like why named exports from CommonJS probably aren't coming soon).

PR #533, opened 2020-07-13
chore: update README.md with `ncu-team sync README.md`

Trott:
Update team list.

Manually re-added Zack Schuster's name.


zackschuster:
is manually re-adding my name every time a hassle? would you prefer if i had it in my profile? i feel bad defeating the automation like this haha

PR #559, opened 2020-10-06
Date  module

harshsri28:
it will represent the  date in this format Tuesday, October 6.

we can call it by
const date = require(__dirname + "/date");

### DIFF
--- a/date.js
+++ b/date.js
@@ -1,0 +1,15 @@
+
+module.exports = getDate;
+
+function getDate(){
+  let today = new Date();
+
+  let options = {
+    weekday: "long",
+    day: "numeric",
+    month: "long"
+  }
+
+  let day = today.toLocaleDateString("en-US", options);
+  return day;
+}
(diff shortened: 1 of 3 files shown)


jkrems:
Hi @harshsri28! Welcome to the world of open source JavaScript! This repository contains notes from discussions around how the module system in nodejs should work ([see: docs on modules](https://nodejs.org/dist/latest-v14.x/docs/api/packages.html)).

If you want to share your JavaScript code with others, you can do so using `npm`. Here's an [example guide for publishing](https://www.digitalocean.com/community/tutorials/workflow-publishing-first-package-to-npm). Each module (or "package") is usually put into its own repository.

ljharb:
This seems like Hacktoberfest spam.

harshsri28:
no it's not spam , as i  add date module

jkrems:
> no it's not spam , as i add date module

The difference between spam and non-spam is if the thing is of value to the recipient and/or was solicited. In this case, adding a generic date utility to a repository for meeting notes isn't a relevant (or valuable) contribution. So from our perspective, it is spam.

harshsri28:
Sorry, sir

harshsri28:
i got it sir

